PR #3, opened 2018-10-10
Sets up stitching

l2succes:
This PR contains the code that outputs the SDL (graphQL Schema) for kaws named `_schema.grapqhl`. This file will then go into metaphysics which we'll use to support the following query

```graphql
{
  collection(id: "andy-warhol") {
    id
    title
    description
    artworks // stitched field
  }
}
```

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1,0 +1,72 @@
+# Object representing a collection page
+type Collection {
+  id: ID!
+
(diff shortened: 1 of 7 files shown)


orta:
🆗

orta:
You might want to make a call on whether you want to keep the changeling too

PR #4, opened 2018-10-11
Make all properties in CollectionQuery nullable

l2succes:
This makes all query properties `nullable`


package.json
```diff
@@ -44,7 +45,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged",
+      "pre-commit": "lint-staged; yarn dump-schema _schema.graphql",
       "pre-push": "yarn build"
```
orta: 👍

PR #12, opened 2018-10-16
Add data dog to kaws

xtina-starr:
Addresses https://artsyproduct.atlassian.net/browse/GROW-943

This follows force's approach to setting up data dog. I'm not really sure how to go about testing it locally. Unless there are other options I figure we could deploy to staging and see if we start seeing things in the datadog dashboard.

It's also worth noting I went ahead and added `DD_APM_ENABLED=true` for both staging and production in hokusai.


eessex:
How do you get the Jira-synced label?

xtina-starr:
@eessex I believe this is a part of the [work](https://artsy.slack.com/archives/C02BC3HEJ/p1539700635000100) that @orta did so I'm assuming by adding a reference to the jira ticket it happens automatically.

src/lib/setup.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const setupDataDog = () => {
+  if (process.env.DD_APM_ENABLED) {
+    ddTracer.init({
+      hostname: process.env.DD_TRACE_AGENT_HOSTNAME,
+      service: "kaws",
+      plugins: false,
+    })
+    ddTracer.use("express", {
+      // We want the root spans of MP to be labelled as just `service`
```
joeyAghion: This comment and the one below are possibly copied from Metaphysics.

xtina-starr: That would make sense. I wasn't sure if they were relevant. I'll remove them.

src/lib/setup.ts
```diff
@@ -0,0 +1,21 @@
+import ddTracer from "dd-trace"
+
+export const setupDataDog = () => {
+  if (process.env.DD_APM_ENABLED) {
```
l2succes: we have to create ENV variables (`DD_APM_ENABLED` and `DD_TRACE_AGENT_HOSTNAME`) both on staging and production before merging this one?

l2succes: @xtina-starr we should also add them to the `.env.example` file

xtina-starr: Does `DD_TRACE_AGENT_HOSTNAME` not get picked up from hokusai [here](https://github.com/artsy/kaws/blob/452936f09b54b351905a5cbde65d7948c36f3b98/hokusai/staging.yml#L24)?

l2succes: My bad, I missed that, it should actually. I'll merge and see if we get any issues

PR #15, opened 2018-10-17
Fix datadog

l2succes:
- Fixes datadog import


src/lib/setup.ts
```diff
@@ -1,4 +1,4 @@
-import ddTracer from "dd-trace"
+import * as ddTracer from "dd-trace"
```
xtina-starr: Hmm wonder why this didn't display an error by the editor. Good catch though.

xtina-starr:
Merge on green

src/lib/setup.ts
l2succes: it's because we're not using webpack. There's an option (`allowSyntheticDefaultImports`) in tsconfig.json that allows default imports in typescript but it doesn't fix the compiled JS import

https://www.typescriptlang.org/docs/handbook/compiler-options.html

PR #16, opened 2018-10-17
Update README

yuki24:
Just making the README more helpful.

#trivial


README.md
```diff
@@ -7,20 +7,67 @@ Named after [the artist](https://artsy.net/artist/kaws), kaws is a backend servi
...
-## How do I work on this?
+## Prerequisites
 
-```sh
-git clone https://github.com/artsy/kaws.git
-cd kaws
-yarn install
+ * node.js 8.12.0 or newer
+ * yarn 1.10.1 or newer
+ * MongoDB 4.x
```
yuki24: I just picked up the latest versions but let me know if we are targeting something else.

l2succes: Yup that looks good

l2succes:
Awesome. Thanks for this @yuki24. A lot better than what we had before!
We can also include links to kubernetes and datadog also at some point

PR #25, opened 2018-10-22
Initialize DataDog tracer before loading other modules

xtina-starr:
Related to: https://artsyproduct.atlassian.net/browse/GROW-943

Per the [documentation](https://github.com/DataDog/dd-trace-js/blob/f638dd66845806ed3ee06e8fbf32b0062b9d21d7/src/proxy.js#L33), we need to make sure we initialize the tracer before loading other libraries in order to correctly setup logging to DataDog.

A todo item we may want to circle back to is ensuring that we can verify what will be sent to DataDog by passing `debug: true` in the options.  For now, we were able to work around this by setting a breakpoint in the [writer.js](https://github.com/DataDog/dd-trace-js/blob/f638dd66845806ed3ee06e8fbf32b0062b9d21d7/src/writer.js#L68) file within the dd-trace library.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,2 +1,0 @@
-# TS build files
-build
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,0 +1,2 @@
+# TS build files
+build
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "dump-schema": "ts-node scripts/dumpSchema.ts",
     "lint": "tslint 'src/**/*.{ts,tsx}'",
     "release": "release-it",
-    "start": "node -r build/index.js",
+    "start": "node build/index.js",
     "test": "jest",
     "type-check": "tsc --noEmit",
     "watch": "tsc --watch"
@@ -58,6 +58,7 @@
     "typeorm": "^0.2.7"
   },
   "devDependencies": {
+    "@types/dd-trace": "^0.6.1",
     "@types/jest": "^23.3.2",
     "@types/lodash": "^4.14.116",
     "@types/mongodb-uri": "^0.9.0",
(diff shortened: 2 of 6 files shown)


src/index.ts
```diff
@@ -1,9 +1,44 @@
-// tslint:disable:no-console
-
+import * as ddTracer from "dd-trace"
```
damassi: Maybe move this line below the `dotenv/config`? That way we can be sure that the first thing being set is our ENV variables.

.vscodeignore
```diff
@@ -0,0 +1,2 @@
+# TS build files
+build
```
damassi: The place where this is more commonly set is in the [`.vscode/settings.json`](https://github.com/artsy/kaws/blob/master/.vscode/settings.json#L6-L12) file.

package.json
```diff
@@ -14,7 +14,7 @@
     "dump-schema": "ts-node scripts/dumpSchema.ts",
     "lint": "tslint 'src/**/*.{ts,tsx}'",
     "release": "release-it",
-    "start": "node -r build/index.js",
```
damassi: 👍 sorry bout that

src/index.ts
xtina-starr: Yeah that makes sense.

.vscodeignore
damassi: Yup, just so that all of our config is consolidated in one place. Here's how we have it [set in reaction](https://github.com/artsy/reaction/blob/master/.vscode/settings.json#L2-L11).

damassi: Which brings up the thought -- in Reaction and elsewhere we use `dist` as the folder, where here we use `build` -- it's not a big deal but the more alignment we have the better. Thoughts @l2succes? (Either way, not something for this pr)

alloy: Yeah good point @damassi, I don’t even think this file worked for us when we added it.

PR #28, opened 2018-10-31
[WIP] add tests for the resolvers

xtina-starr:
Addresses: https://artsyproduct.atlassian.net/browse/GROW-937

### DIFF
--- a/src/Resolvers/Collections.ts
+++ b/src/Resolvers/Collections.ts
@@ -4,7 +4,7 @@ import { Collection } from "../Entities/Collection"
 
 @Resolver(of => Collection)
 export class CollectionsResolver {
-  private readonly repository = getMongoRepository(Collection)
+  protected readonly repository = getMongoRepository(Collection)
 
   // TODO: should return a connection
   @Query(returns => [Collection])
--- a/src/__tests__/data.ts
+++ b/src/__tests__/data.ts
@@ -1,0 +1,41 @@
+import { plainToClass } from "class-transformer"
+// import { MongoRepository, EntityRepository } from "typeorm"
+import { Collection } from "../Entities/Collection"
+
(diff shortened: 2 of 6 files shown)


src/Resolvers/Collections.ts
```diff
@@ -4,7 +4,7 @@ import { Collection } from "../Entities/Collection"
 
 @Resolver(of => Collection)
 export class CollectionsResolver {
-  private readonly repository = getMongoRepository(Collection)
+  protected readonly repository = getMongoRepository(Collection)
```
yuki24: Do we actually need this change?

src/__tests__/data.ts
```diff
@@ -0,0 +1,41 @@
...
+    headerImage:
+      "http://files.artsy.net/images/pumpkinsbigartistsmallsculpture.png",
+    query: {
+      id: null,
+      gene_id: null,
+    },
+  },
+])
+
+// @EntityRepository()
```
l2succes: Let's remove the commented code

src/Resolvers/Collections.ts
l2succes: Actually no, we don't need this anymore as it was meant to expose the `repository` to a child class for mocking purposes

l2succes:
Also we need to move the fixtures data out of `__tests__/` because jest is picking it up
https://circleci.com/gh/artsy/kaws/121

package.json
```diff
@@ -21,6 +21,7 @@
   },
   "jest": {
     "preset": "ts-jest",
+    "testMatch": ["**/__tests__/**/*test.ts"],
```
xtina-starr: @l2succes versus move the fixture data file out of the test directory, I added this to ignore files that weren't tests.

PR #30, opened 2018-11-06
Group collections by category

eessex:
- Adds ability to query `collections` and receive collections grouped by category names
- Fixes construction of mongo connection args to work in local environments


src/index.ts
```diff
@@ -48,23 +48,40 @@ import { Connection, createConnection } from "typeorm"
 import { createSchema } from "./createSchema"
 import { entities } from "./Entities"
 
-bootstrap()
+export const mongoConnectArgs = () => {
```
l2succes: let's extract that to a `config/database.ts` module and name it `config`

PR #33, opened 2018-11-14
Adds a 'yarn dev' command

### LINKED ISSUE #27: Improve dev server start UX
After pairing with @xtina-starr on some kaws work yesterday, a few pieces of feedback came up:

* Currently one needs to remember to run `yarn build` before running `yarn start`, which is too easy to forget. One should really only need to run `yarn start` and be up and running.
* Because kaws has server code, you may want to leverage a start script, like [the one in Force](https://github.com/artsy/force/blob/76ee36087046661900f068c64e492d0d7bf0c871/scripts/start.sh#L19-L21), so that when running from vscode’s terminal you automatically get the debugger connected.

orta:
Adds a dev command to fix #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,30 @@
-kaws
-===
+# kaws
 
-Named after [the artist](https://artsy.net/artist/kaws), kaws is a backend service that powers [artsy.net](https://artsy.net) collection pages. What are collections you ask? A Collection is a prefiltered version of Artsy's [Collect](https://artsy.net/collect) page for marketing purposes.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -26,35 +34,38 @@ First, make sure to install dependencies:
...
-> db.
+> As of writing, this command doesn't exit automatically. You can kill the
+> process once you confirm the data is in the db.
 
 Finally, start the server by running the command below:
 
 ```bash
-yarn start
+yarn dev
 ```
```
orta: this is the only real change

l2succes: 👍

PR #36, opened 2018-11-26
@l2succes => Pass thru `marketable` to filter artworks

mzikherman:
Think we need this?


mzikherman:
Ok, this might not be needed!

l2succes:
Since we may need to support overriding the `marketable` flag on a given collection. This actually makes sense after all.

PR #40, opened 2018-12-17
Update Kaws' CSV import script to also update the sitemap seed JSON

yuki24:
finishes https://artsyproduct.atlassian.net/browse/GROW-1051

 * Adds `aws-sdk` dependency
 * Extracts `convertCSVToJson` into its own file so `scripts/bootstrapOrUpdate.ts` and `scripts/updateSitemapJson.ts` share the logic
 * Adds `yarn update-sitemap ./collections.csv` that updates the S3 object

## Todo

 * [ ] Add tests?

### DIFF
--- a/scripts/updateSitemapJson.ts
+++ b/scripts/updateSitemapJson.ts
@@ -1,0 +1,64 @@
+import "dotenv/config"
+
+import { Collection } from "../src/Entities"
+import { loadCsvAsCollection } from "./loadCsvAsColleciton"
+
+const AWS = require("aws-sdk")
+
+const S3_COLLECT_BUCKET = "artsy-data"
+const S3_COLLECT_KEY = "collect/collect_and_collections.json"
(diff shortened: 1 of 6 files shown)


scripts/updateSitemapJson.ts
```diff
@@ -0,0 +1,64 @@
+import "dotenv/config"
+
+import { Collection } from "../src/Entities"
+import { loadCsvAsCollection } from "./loadCsvAsColleciton"
+
+const AWS = require("aws-sdk")
```
yuki24: I'm not sure why not `import` did not work but `require` did...

l2succes: did you try `import * as AWS from "aws-sdk"`

yuki24:
Closing as I'll have to re-write the entire script to use MongoDB.

PR #41, opened 2018-12-17
moves convertCSVToJSON to separate module

l2succes:
- fixes a bug with the csv to json conversion function
- adds tests


yuki24:
The peril comment seems out-of-date so I'll just go ahead and merge.

PR #42, opened 2018-12-17
Add a new script that generates a new sitemap based on collections in the DB

yuki24:
finishes https://artsyproduct.atlassian.net/browse/GROW-1051

This is a re-write of #40 but this PR adds a script that relies on the database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,17 +45,15 @@ cp .env.example .env
 Usually you do not have to change them, but if you need, update the variables to
 match your local development setup.
 
-Once this is done, the next step would be to load test data. This repository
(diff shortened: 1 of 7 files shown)


yuki24:
I think the `aws-sdk` dependency is reasonable here. Also I forgot to update my own test but it should be good to go.

yuki24:
The build is fixed and now this is ready for review.

yuki24:
The test has been fixed by 0ddb7e1ceec565e5e3d403e9040e53a14e72cc4a so I'll go ahead and merge this PR. Thanks!

PR #43, opened 2019-01-03
Update sitemap as part of bootstrap-data

yuki24:
We seem to have fotgotten to run `yarn update-sitemap` after a batch update so let's just run it as part of `yarn bootstrap-data` so we never forget to run it (or do forget about it and not worry about it).

#trivial

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   ],
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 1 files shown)


yuki24:
I''' go ahead and merge this since we've forgotten this twice in a month and I would like this in as soon as possible.

PR #44, opened 2019-01-15
Avoid group mongodb function since it's not supported on lower-end Atlas clusters

joeyAghion:
After migrating Kaws's staging database (as part of PLATFORM-1079), we noticed staging's `/collections` page failing with a server-side error. The underlying graphql query was failing with `Error: CMD_NOT_ALLOWED: group`.

I couldn't find clear documentation of this in Atlas's [list of unsupported commands](https://docs.atlas.mongodb.com/reference/free-shared-limitations/#atlas-free-tier), but maybe `group` actually relies on the aggregation framework in v4. This switches the query to use a `distinct`, with additional queries to load the collections for each category.

### DIFF
--- a/src/Resolvers/Collections.ts
+++ b/src/Resolvers/Collections.ts
@@ -16,27 +16,12 @@ export class CollectionsResolver {
   // TODO: should return a connection
   @Query(returns => [CollectionCategory])
   async categories(): Promise<CollectionCategory[]> {
-    const data = await this.repository.group(
-      { category: 1 },
-      { category: { $exists: true } },
-      {},
-      (curr, result) => {
-        if (!result.collections) {
-          result.collections = [curr]
-        } else {
-          result.collections.push(curr)
-        }
-      },
-      result => result,
-      true
-    )
-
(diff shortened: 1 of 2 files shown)


src/Resolvers/Collections.ts
```diff
@@ -16,27 +16,14 @@ export class CollectionsResolver {
...
+    })
+    return categories.map(category => ({
       name: category,
-      collections: collections.map(col => ({
-        ...col,
-        id: col._id,
-      })),
+      collections: this.repository
+        .find({ category })
+        .then(cols => cols.map(col => ({ ...col }))),
```
l2succes: I'm not sure I understand why that map is required here. Can you explain?

joeyAghion: I think you're right. At first I thought we'd have to transform the data to do things like the `id: col._id` above. But from what I can tell, that's not necessary. Pushing...

l2succes: Got it, yup that makes sense 👍

PR #48, opened 2019-01-29
@dblandin => [Search] Adds task to index collections data into ElasticSearch

mzikherman:
@l2succes 

This adds a task that will index all collections data into a configured ElasticSearch instance. Locally it took ~2 seconds to run and index all collections (pointing to staging's Mongo instance).

```
➜  kaws git:(index_to_elasticsearch) curl -XGET 'localhost:9200/marketing_collections_development/marketing_collection/5beb5aca1338c043ac8ae214?pretty'
{
  "_index" : "marketing_collections_development",
  "_type" : "marketing_collection",
  "_id" : "5beb5aca1338c043ac8ae214",
  "_version" : 1,
  "found" : true,
  "_source" : {
    "name" : "KAWS: Toys",
    "keywords" : "Companion, BFF, Passing Through, Astroboy, Astro Boy, Small Lie, Together, Accomplice, Twins, Stormtrooper, Boba Fett, Darth Vader, Undercover Bear, Tweety, Pinocchio, Resting Place, Along the Way, Final Days, Zooth, Tweety, Joe Kaws, Snoopy, Seeing Watching, Pinocchio, Jiminy Cricket, Partners, Milo, Kubrick, Zooth, Bounty Hunter, Bearbrick, Chum, JPP, Final Days, Chompers, Cat Teeth Bank, Bend, Blitz, Bendy, Accomplice"
  }
}
```

Since the data is so small, I think a task that re-indexes everything (possibly on a cron, or just manually kicked off once data is re-imported) is sufficient.

### Migration

- Add `ELASTICSEARCH_URL` to staging/prod
- Add Gravity PR that includes `MarketingCollection` as a search model w/ schema, allow searching
- Re-index KAWS data
- ...success! At this point, the data will be present in ES and Gravity `/match` endpoints should be returning data.


src/utils/indexForSearch.ts
```diff
@@ -0,0 +1,42 @@
...
+      const repository = getMongoRepository(Collection)
+
+      const collections = await repository.find()
+      for (const collection of collections) {
+        const keywords = collection.query.keyword
+        await search.client.index({
+          index: search.index,
+          type: "marketing_collection",
+          id: collection.id.toString(),
+          body: { name: collection.title, keywords },
```
mzikherman: This schema will probably change. It seems as though Gravity's querying for ES will use `name`, `featured_names`, and `alternate_names` as the fields. So I imagine the keywords here should actually be placed into `alternate_names`.

src/lib/search.ts
```diff
@@ -0,0 +1,14 @@
+const { NODE_ENV, ELASTICSEARCH_URL } = process.env
+const elasticsearch = require("elasticsearch")
+
+const client = new elasticsearch.Client({
```
l2succes: consider adding a check if url is missing or just double check the script exits gracefully in that instance

src/utils/indexForSearch.ts
l2succes: Aside from keywords. There are more properties you can add to the collection data payload that'll come in handy. Namely `query.artist_ids`, `query.tag_id`, `query.gene_ids`. This will allow us to do more complex querying such as the use case you asked about: looking up collections by `artist_id`.

mzikherman: So, I'm not sure we need that...those are just using ElasticSearch as a giant denormalized data store (not the ideal use case, we do this a lot and it's an anti pattern).

Instead, I'm _just_ adding the text fields which will actually be using the text analysis + search functionality.

To answer questions like: what genes/artists/tags have associated collections, we should just allow querying KAWS with those fields, and add `collections` into those types w/ a `delegateSchema` call to KAWS.

l2succes: Right... that makes sense then. So maybe just consider `collection.category` (e.g. "Abstract Art")  which is also a text field ES can consider a search term & users could search collections by

mzikherman: great call, will add that!

mzikherman:
Alright updated! I think this might be good to go, it all runs locally. I'll set `ELASTICSEARCH_URL` on staging and prod.

I'll run it via: `hokusai staging run 'yarn index-for-search' --tty` (🤞 ). Later on can make this a cron, or hook it up into the process which updates collections, etc.

mzikherman:
Ran this in staging...looks good.

PR #53, opened 2019-01-31
@l2succes => [Collections] Allow querying for collections by artist id

mzikherman:
Tested locally against staging, seems to work!

I skipped tests since they wouldn't add anything in this case: https://github.com/artsy/kaws/blob/9834452bc011903ad90245467d54f068c131b0ff/src/Resolvers/__tests__/collection_resolvers.test.ts#L11-L12 , everything is stubbed. Although, a spec that at least checks that the schema accepts a query like `collections(artistID: "...")` might be useful. It wouldn't really exercise whether the lookup functionally works, but it would assert something about the shape of the schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### Master
 
+- Return collections for an artist - [@mzikherman]
 - Enable text search - [@mzikherman]
 - Initial setup - [@l2succes]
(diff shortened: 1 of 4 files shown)


mzikherman:
Merging! (I think this is pretty innocuous, but MP is ready to go).

PR #55, opened 2019-02-01
Add a missing MONGOHQ_URL to the command that updates sitemap seed

yuki24:
This is missing in the README and is required to generate the sitemap seed with the production database. If it's run without it it'll default to the local database which may or may not have the same data as production and we may end up generating an almost empty sitemap.

#trivial

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,17 +75,6 @@ query {
 }
 ```
 
-## Syncing Kaws' database with the data in Google Spreadsheet
-
-When we update the production database with the `update-database` command, we
-also need to inform Google of the updates so newly added collections will be
-crawled. You will need your `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to
-update an object in S3 to run the command below:
-
-```bash
-AWS_ACCESS_KEY_ID=... AWS_SECRET_ACCESS_KEY=... yarn update-sitemap
-```
-
 ## Testing
 
 ```sh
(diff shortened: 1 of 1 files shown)


joeyAghion:
Should this be run via `hokusai production run '...'` instead, so it's consistent and not sensitive to developers' local environments and config?

yuki24:
@joeyAghion Ideally yes. I wasn't sure how I could do this in a JS project (though it's probably just wrapping the command as you presented).

yuki24:
We may have to add `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to the production pods if we want to go with the hokusai path.

iskounen:
I was thinking about moving this section in the README to a playbook in potential so that developers don't unwittingly run this thread while setting up their dev environment (as I did...). I'm curious what you guys would think about that?

joeyAghion:
@yuki24 yup I think it would actually be a good thing to add those to Kaws's config map (based on a properly scoped "machine" IAM profile) if that privilege is necessary for this system to operate correctly. I think it would literally be `hokusai production run 'yarn update-sitemap'` in that case.

@iskounen in general I like docs about a particular system to stay within the repo (this is actually [in a playbook](https://github.com/artsy/README/blob/master/playbooks/documentation.md#documentation-or-discussion-of-a-specific-system)). Could we safeguard it another way, or maybe put it in a separate doc about production operations?

mdole:
yep I think Joey is correct on this one about the command to use, and I also like the idea of keeping it to hokusai production. also +1 to making the doc as close to KAWS as possible—is there a reason we couldn't just include something in the README like "Artsy staff with IAM credentials can use `hokusai production run 'yarn-update-sitemap'` to trigger a sitemap update in S3"? too sensitive?

yuki24:
I've updated the README to wrap the command with `hokusai run ...`.  I'm also making this PR as WIP as we have to generate and set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` in production.

I'm not sure if this will make a difference, but I've moved the `Updating the sitemap for Collect` to the end to make it not look like the part of Getting started.

yuki24:
It seems like there's a user called `kaws` in our IAM, so I just added a new policy that gives the user read/write access to the `artsy-data` policy. Also both `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are now in production.

yuki24:
I'm going to merge this PR if there are no more objections. Any thoughts?

mdole:
seems good to me!

PR #57, opened 2019-02-01
[WIP] Send Collections to Sailthru

mdole:
This PR implements a rudimentary script to sync collections to Sailthru as per [this ticket](https://artsyproduct.atlassian.net/browse/GROW-958)

As it stands, if you update the Sailthru Key + Secret and the Mongo URL in the .env file and run it, this would sync collections to Sailthru. However, there are some things that would be good to resolve/improve:

- [ ] Some slugs need to be sanitized. For example: 
    - KAWS slug: john-baldessari-raised-eyebrows/furrowed-foreheads
    - Actual slug: john-baldessari-raised-eyebrows-furrowed-foreheads
- [ ] Add artworks included in collection
- [ ] Add artists included in collection
- [ ] Tests, tests, tests

Not assigning any reviewers just yet since there's still some work to be done before this gets merged!


src/utils/getArtworks.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (tag_id) {
+    subQuery.push({
+      term: {
+        tags: tag_id,
+      },
+    })
+  }
+
+  const response = await search.client.search({
+    index: `artworks_staging`,
```
mzikherman: The name of the artworks index is actually `gravity_application_staging/production` (that's a legacy thing from when we only had artworks in ES and so didn't have a better index name specified)

mdole: thanks for the note @mzikherman !

l2succes: @mzikherman I checked our ES cluster config & it says `artworks_staging` is an alias for `gravity_application_staging` so they point to the same index? and the query works

mzikherman: Ah, I didn't know we had that setup? In that case 👌

mdole:
@l2succes mind taking a quick look at this? going to do a separate PR with the artist/artwork work, so would like to get this merged so that I can wrap up with that

PR #58, opened 2019-02-06
Price guidance field

eessex:
Adds price guidance field to Collection entity

Addresses https://artsyproduct.atlassian.net/browse/GROW-1107


kaws.graphql
```diff
@@ -0,0 +1,80 @@
+# Object representing a collection page
+type Collection {
```
eessex: Is this file supposed to be merged, or does it live in `.gitignore`?

src/Entities/Collection.ts
```diff
@@ -66,4 +66,11 @@ export class Collection {
   @Field()
   @Column()
   updatedAt: Date
+
+  @Field(type => Number, {
+    nullable: true,
+    description: "Suggested average price for included works",
+  })
+  @Column()
```
l2succes: `@Column({ nullable: true })`

PR #59, opened 2019-02-06
Add show_on_editorial boolean field to collections

eessex:
Adds a new boolean field `show_on_editorial` to the collection entity.

This unblocks us for syncing new spreadsheet data into the DB, I'll continue to work on queries in a separate PR.


src/utils/convertCSVToJSON.ts
```diff
@@ -39,6 +40,7 @@ export const convertCSVToJSON: (string) => Promise<Collection[]> = (
                 headerImage,
                 credit,
                 price_guidance,
+                show_on_editorial,
```
l2succes: we should add default value to `false` here & make sure we test ingestion script

eessex: Thanks for the reminder, this is now updated and tests as well.

PR #61, opened 2019-02-08
[WIP] Querying metaphysics for artist and artwork data

mdole:
This adds some new functions and updates the update-sailthru script to include artist and artwork data in what's being sent to Sailthru. This can be merged as it's just a script update, but I'd like to see if I can figure out a test or two first.

Also: this currently does things in a pretty inefficient way, as we're querying Metaphysics twice for each collection being pushed to Sailthru. We could get that down to one pretty easily, but as discussed with @l2succes, even better would be to actually add artist and artwork information to each collection in MongoDB.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,5 +5,16 @@ DD_APM_ENABLED=false
 SAILTHRU_KEY=
 SAILTHRU_SECRET=
 
+# Needed to push content to Sailthru - different versions exist for staging and
+# production instances!
+SAILTHRU_KEY=
+SAILTHRU_SECRET=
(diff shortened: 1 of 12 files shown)


mdole:
@l2succes do you think it makes sense to merge this as-is and make a separate ticket for improvements, or just leave it open and have someone on Grow take the final steps when they have time?

l2succes:
Nice work @mdole! Yes, Let's go ahead and merge this one. I'll create a separate ticket for the final steps which we can handle on grow.

PR #62, opened 2019-02-13
@pepopowitz => Support `size` for querying collections

mzikherman:
As discussed, this ~~adds `image_url` to the search index and~~ supports passing in a size param.

Since this gets stitched into the Metaphysics schema and this changes the KAWS schema, there's a small MP PR needed to update the stored schema there, as well as pass this arg along.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -77,7 +77,11 @@ type CollectionQuery {
 scalar DateTime
 
 type Query {
-  collections(showOnEditorial: Boolean, artistID: String): [Collection!]!
+  collections(
+    size: Int
+    showOnEditorial: Boolean
+    artistID: String
+  ): [Collection!]!
   categories: [CollectionCategory!]!
   collection(slug: String!): Collection
 }
--- a/src/Resolvers/Collections.ts
+++ b/src/Resolvers/Collections.ts
@@ -1,4 +1,4 @@
-import { Arg, Query, Resolver } from "type-graphql"
+import { Arg, Int, Query, Resolver } from "type-graphql"
 import { getMongoRepository } from "typeorm"
 import { Collection } from "../Entities/Collection"
 import { CollectionCategory } from "../Entities/CollectionCategory"
@@ -10,7 +10,8 @@ export class CollectionsResolver {
   @Query(returns => [Collection])
   async collections(
     @Arg("artistID", { nullable: true }) artistID: string,
-    @Arg("showOnEditorial", { nullable: true }) showOnEditorial: boolean
+    @Arg("showOnEditorial", { nullable: true }) showOnEditorial: boolean,
+    @Arg("size", () => Int, { nullable: true }) size: number
   ): Promise<Collection[]> {
     const hasArguments =
       [].filter.call(arguments, arg => arg !== undefined).length > 0
@@ -21,6 +22,9 @@ export class CollectionsResolver {
     }
     if (artistID) {
       query.where["query.artist_ids"] = { $in: [artistID] }
+    }
+    if (size) {
+      query.take = size
     }
     return await this.repository.find(query)
   }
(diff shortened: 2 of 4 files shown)


_schema.graphql
```diff
@@ -77,7 +77,11 @@ type CollectionQuery {
 scalar DateTime
 
 type Query {
-  collections(showOnEditorial: Boolean, artistID: String): [Collection!]!
+  collections(
+    size: Float
```
alloy: As commented on the related MP PR, should this be an integer?

mzikherman: Probably, but when the schema was generated here it created `Float`. That comes from here: https://github.com/artsy/kaws/pull/62/files#diff-500113cce4e1c0ce1050b25c61eb35a4R14

`size: number`

Was not able to figure out with this library how to change that? Since it gets derived from the TS typings (where there is no integer, right?).

Overall this library was a bit of a challenge to use and figure out. I get what you're saying...but I _think_ this is fine (plus couldn't figure out anything else...so would love another pair of eyes if you're volunteering!). Otherwise we'll roll with this so as not to block on a type thing. We can always change it later.

orta: Yeah, looks like the library must prefer float representation - it does that here also. https://github.com/artsy/kaws/blob/master/_schema.graphql#L31-L32

alloy: Well, that’s not great… searching [the repo](https://github.com/19majkel94/type-graphql) I can’t even find a single mention about this.

I'm not of the opinion that this PR should be blocked, but I do think that long term the schema should be self-documenting and not be confusing due to some JS implementation details (correct, there is no integer JS type). The change from `Float` to `Int` will technically be a breaking one, but as long as the app isn't using this it should probably be fine (although it will probably require a concerted effort).

@l2succes Did you run into this yet and consider it?

eessex: I ran into a similar issue on `price_guidance`, which resulted in a bug when added to Metaphysics -- https://artsyproduct.atlassian.net/browse/GROW-1126

This field is supposed to be an optional `number`, but the schema labels it as a `Float`.  When empty, it results in the error that a float cannot be empty.  Not sure if this is directly related, but an unexpected result of the typecasting from this library.

l2succes: Yes, you can explicitly specify the type of a graphQL argument. The following way
```typescript
import { Arg, Int, Query, Resolver } from "type-graphql"

@Query(returns => [Collection])
  async collections(
    ...
    @Arg("size", () => Int, { nullable: true }) size: number
  ): Promise<Collection[]> {
  ...
}
```


This will generate the following SDL.
```graphql
type Query {
  collections(
    size: Int
  ): [Collection!]!
}
```

`type-graphql` actually has really good documentation, you can read more about how to set types here
https://19majkel94.github.io/type-graphql/docs/resolvers.html#arguments

Let me know if you have any more questions

src/Resolvers/Collections.ts
```diff
@@ -10,7 +10,8 @@ export class CollectionsResolver {
   @Query(returns => [Collection])
   async collections(
     @Arg("artistID", { nullable: true }) artistID: string,
-    @Arg("showOnEditorial", { nullable: true }) showOnEditorial: boolean
+    @Arg("showOnEditorial", { nullable: true }) showOnEditorial: boolean,
+    @Arg("size", { nullable: true }) size: number
```
l2succes: `@Arg("size", () => Int, { nullable: true }) size: number`

@mzikherman See answer above

_schema.graphql
alloy: Thanks!

src/Resolvers/Collections.ts
mzikherman: Worked great, thanks!

mzikherman:
Ok, merging this- its now an `Int`! cc @eessex you can make the price field int as well now if needed.

PR #69, opened 2019-02-22
update-sailthru QoL improvements

mdole:
The `update-sailthru` script was halting on an error— @l2succes and I paired on finding the error, resolving, and making some other changes. Tweaks made:

- Consolidated down to single call to Metaphysics
- Stopped script from halting on error (prints and continues on)

The error seems to be a result of URLs that have special characters in them, such as https://www.artsy.net/collection/salvador-dalí-venus

In case someone else needs to find and resolve errors with this script in the future, this command runs the script in debugging mode:
`yarn nodemon --inspect-brk -r ts-node/register scripts/pushCollectionsToSailthru.ts`
You might also need to change your METAPHYSICS_URL and MONGOHQ_URL env vars depending on where you plan to query.


mdole:
hmm working on that test

l2succes:
@mdole are you still working on the test?

mdole:
Fixed in the most recent commit! I'm ooo this week; if it looks good to you
feel free to merge but if there's anything I should add I can do so when
I'm back :)

PR #72, opened 2019-02-26
pass the correct query format

xtina-starr:
Fixes QA issues related to [GROW-1145](https://artsyproduct.atlassian.net/browse/GROW-1145)

As I was QA'ing the new `randomize` arg, I noticed that when it was passed with other arguments (i.e. `showOnEditorial: true`) we were returning an empty array. This was because I was passing the incorrect query format. I was previously passing `{ where: { show_on_editorial: true } }` to the match property when it should be ` { show_on_editorial: true }`. This PR fixes that. The work in Kaws related to randomize should be legit now.


eessex:
LGTM!

PR #74, opened 2019-02-28
Add boolean field is_featured_artist_content

eessex:
Related to https://artsyproduct.atlassian.net/browse/GROW-1143

- Adds a boolean field to determine if an artist-related collection is featured in the artist page, and potentially future contexts.
- Adds query to return only collections based on `is_featured_artist_content` value

One q -- I noticed all fields except `headerImage` on the collections model use snake_case, is do we want to standardize these and backfill?


src/utils/convertCSVToJSON.ts
```diff
@@ -41,6 +42,7 @@ export const convertCSVToJSON: (string) => Promise<Collection[]> = (
                 credit,
                 price_guidance: price_guidance ? Number(price_guidance) : null,
                 show_on_editorial: Boolean(show_on_editorial),
+                is_featured_artist_content: Boolean(show_on_editorial),
```
l2succes: why are we using `show_on_editorial` here?

eessex: Ah! Copy-paste, will update this now

src/Entities/Collection.ts
```diff
@@ -79,4 +79,10 @@ export class Collection {
   })
   @Column({ default: false })
   show_on_editorial: boolean = false
+
+  @Field(type => Boolean, {
+    description: "Collection has prioritized connection to artist",
```
l2succes: I'd say let's be a bit more explicit in the description `Collection should be featured on artist page`

eessex: I intentionally made this more vague, so that we indicate that it is a high priority in relationship to its affiliated artist, but potentially leave it open to other contexts where this flag could be relevant.

eessex: For example, we could display a similar rail on an artwork page to indicate other works by an artist, in which case the connection to the artist remains true, but the context of the artist page does not.

l2succes: Got it, sounds good then

PR #75, opened 2019-03-01
Adds sentry to kaws

xtina-starr:
Addresses: [GROW-1120](https://artsyproduct.atlassian.net/browse/GROW-1120)

This adds error tracking for the kaws project.

ToDo:
- [x]  Setup Kaws projects in Sentry for staging and production.
- [ ]  Set the environment variables.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,7 @@ import * as morgan from "morgan"
 import { Connection, createConnection } from "typeorm"
 import { databaseConfig } from "./config/database"
 import { createSchema } from "./utils/createSchema"
-const enableSentry = !!SENTRY_PRIVATE_DSN
+const enableSentry = Boolean(SENTRY_PRIVATE_DSN)
 
 bootstrap()
 
@@ -53,7 +53,7 @@ async function bootstrap() {
     // Setup Sentry
     if (enableSentry) {
       Sentry.init({
-        dsn: "https://caca4ef0651e4adbbd1f099702dbb779@sentry.io/1405985",
+        dsn: SENTRY_PRIVATE_DSN,
       })
 
       // The request handler must be the first middleware on the app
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -10,12 +10,14 @@ setupDataDog()
...
 import "reflect-metadata"
 
+import * as Sentry from "@sentry/node"
 import { GraphQLServer, Options } from "graphql-yoga"
 import { parse } from "mongodb-uri"
 import * as morgan from "morgan"
 import { Connection, createConnection } from "typeorm"
 import { databaseConfig } from "./config/database"
 import { createSchema } from "./utils/createSchema"
+const enableSentry = !!SENTRY_PRIVATE_DSN
```
eessex: what is this `!!` syntax?

l2succes: it's the same as `Boolean(SENTRY_PRIVATE_DSN)` it basically converts an truthy or falsy value to a proper boolean

src/index.ts
```diff
@@ -48,6 +50,19 @@ async function bootstrap() {
       debug: NODE_ENV === "development",
     }
 
+    // Setup Sentry
+    if (enableSentry) {
+      Sentry.init({
+        dsn: "https://caca4ef0651e4adbbd1f099702dbb779@sentry.io/1405985",
```
l2succes: @xtina-starr we should use the env variable for this correct?

xtina-starr: oh shoot I actually set it up that way and for got to remove this

src/index.ts
xtina-starr: `Boolean(SENTRY_PRIVATE_DSN)` might actually be better.

eessex: Agreed that seems more communicative.

src/index.ts
xtina-starr: Good catch @l2succes! This was for the staging environment. I'll go ahead and generate a new key.

xtina-starr:
@eessex I don't think so especially considering there is a conditional check for the presence of the env variable. I just added the ToDo to remind myself.

PR #76, opened 2019-03-01
use randomizationSeed instead of randomize boolean

xtina-starr:
This changes support for randomization to use a `randomizationSeed` that passes a unique string. This is to address the issue of the collections rail component getting updated with the same collection data despite being on different articles. This is caused because the way relay handles rendering component when the same parameters are passed.

ToDo:

- [x] Create a PR in metaphysics with schema changes.
- [x] Create a PR in reaction to sync the schema and use the updated parameter.

### DIFF
--- a/src/Resolvers/Collections.ts
+++ b/src/Resolvers/Collections.ts
@@ -13,7 +13,7 @@ export class CollectionsResolver {
     @Arg("artistID", { nullable: true }) artistID: string,
     @Arg("showOnEditorial", { nullable: true }) showOnEditorial: boolean,
     @Arg("size", () => Int, { nullable: true }) size: number,
-    @Arg("randomize", { nullable: true }) randomize: boolean
+    @Arg("randomizationSeed", { nullable: true }) randomizationSeed: string
   ): Promise<Collection[]> {
     const hasArguments =
       [].filter.call(arguments, arg => arg !== undefined).length > 0
@@ -25,11 +25,11 @@ export class CollectionsResolver {
     if (artistID) {
       query.where["query.artist_ids"] = { $in: [artistID] }
     }
-    if (!randomize && size) {
+    if (!randomizationSeed && size) {
       query.take = size
     }
 
-    if (randomize) {
+    if (randomizationSeed) {
       const aggregatePipeline: any = []
       if (!isEmpty(query.where)) {
         aggregatePipeline.push({ $match: query.where })
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -78,7 +78,7 @@ scalar DateTime
 
 type Query {
   collections(
-    randomize: Boolean
+    randomizationSeed: String
     size: Int
     showOnEditorial: Boolean
     artistID: String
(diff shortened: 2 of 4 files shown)


alloy:
@xtina-starr So did this fix the issue?

Also, as a side-note, currently kaws doesn’t actually do anything with the seed, but to complete the API the seed should actually be used in a way that results in the same set of collections being returned.

alloy:
Hrmm, doesn’t look like MongoDB actually supports providing your own random seed :( https://jira.mongodb.org/plugins/servlet/mobile#issue/SERVER-22225

orta:
My gut take says the underlaying issue is still a mis-understanding of Relay's caching, but now that queries for each random collection is unique (because it has the string of the article it in) then all instances of the random collection doesn't get updated at the same time

alloy:
Ace 👌

I don’t think it’s a misunderstanding, I think it’s reasonable for Relay to assume that the graph is deterministic. I.e. the same input variables must refer to the same data set, so if that data changes in the store any live components that refer to it will get updated with the new data.

I wonder if there’s another way to influence that store “selector” (is what I think Relay calls it internally).

l2succes:
That makes sense, that was my assumption as well. 

I was hoping there was a way with Relay to pass in an arbitrary (client-side only) parameter into a GraphQL query without actually changing the underlying schema since the seed is only being used to differentiate subsequent queries here.

src/Resolvers/Collections.ts
```diff
@@ -13,7 +13,7 @@ export class CollectionsResolver {
     @Arg("artistID", { nullable: true }) artistID: string,
     @Arg("showOnEditorial", { nullable: true }) showOnEditorial: boolean,
     @Arg("size", () => Int, { nullable: true }) size: number,
-    @Arg("randomize", { nullable: true }) randomize: boolean
+    @Arg("randomizationSeed", { nullable: true }) randomizationSeed: string
```
l2succes: In terms of the API, could we actually go with the following 
```graphql
collections(randomize: true, seed: "arbitrary_string") {
  id
}
```
as I think it makes the intent clearer to the consumer. We can also add a description to those arguments

e.g.
```
@Arg("randomize", () => Boolean, { nullable: true, description: "Return a random sample of collection. Must specify size and seed" }) randomize: boolean
```

orta:
Yeah, I was thinking a mis-understanding from our side. I have two other ideas:

- Adding extra (unused?) variables to the query from the client, but I'm willing to bet it won't get past the relay compiler. 

- I wondered if adding Node conformance to collections from Kaws would also fix the underlaying issue

xtina-starr:
@l2succes, I believe @alloy and I discussed passing the randomize boolean and seed and felt that `randomizationSeed` sort of self documents and prevents the need to pass or couple two arguments.

xtina-starr:
@l2succes @alloy @orta this blocks an A/B test that we are trying to get out today. Are there strong opinions against the existing implementation?

It's also worth noting that we can continue the conversation and use the closing of the test as an opportunity to implement whatever better solution we come up with. I just want to keep movement on this seeing that it's been held up for a while.

l2succes:
Got it, thanks for the context @xtina-starr. let's go for it then.

PR #78, opened 2019-03-07
[HOLD] Save featured artworks to a collection

eessex:
This gives us the functionality needed for the current rail design, will continue spiking a cache strategy to get these or other gravity data from metaphysics instead.

- Adds field `featuredArtworks` array to collection schema
- Adds functionality to fetch top-3 artworks by merchandisability
- Adds yarn task to fetch `featuredArtworks` and save for all collections

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -1,3 +1,13 @@
+type ArtworkImage {
+  aspectRatio: Int
+  height: Int
+  width: Int
+  position: Int
+  url: String
+  imageUrl: String
+  versions: [String!]
+}
+
 # Object representing a collection page
 type Collection {
   id: ID!
@@ -51,17 +61,7 @@ type CollectionFeaturedArtwork {
   id: String
 
   # Artwork image data
-  image: CollectionFeaturedArtworkImage!
-}
-
-type CollectionFeaturedArtworkImage {
-  aspectRatio: Int
-  height: Int
-  width: Int
-  position: Int
-  url: String
-  imageUrl: String
-  versions: [String!]
+  image: ArtworkImage!
 }
 
 type CollectionQuery {
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -36,11 +36,32 @@ type Collection {
 
   # Collection has prioritized connection to artist
   is_featured_artist_content: Boolean!
+
+  # Array of top merchandisabile artworks and associated image
+  featuredArtworks: [CollectionFeaturedArtwork!]
 }
 
 type CollectionCategory {
   name: String!
   collections: [Collection!]!
+}
+
+type CollectionFeaturedArtwork {
+  _id: ID
+  id: String
+
+  # Artwork image data
+  image: CollectionFeaturedArtworkImage!
+}
+
+type CollectionFeaturedArtworkImage {
+  aspectRatio: Int
+  height: Int
+  width: Int
+  position: Int
+  url: String
+  image_url: String
+  versions: [String!]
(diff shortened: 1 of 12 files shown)


_schema.graphql
```diff
@@ -36,13 +36,34 @@ type Collection {
...
+  image: CollectionFeaturedArtworkImage!
+}
+
+type CollectionFeaturedArtworkImage {
+  aspectRatio: Int
+  height: Int
+  width: Int
+  position: Int
+  url: String
+  imageUrl: String
```
eessex: Added these as camelCase, but we may want to reconsider because we are manually renaming the metaphysics return values for `aspect_ratio` and `image_url`.

l2succes: It makes sense to do so as that's what we're currently doing for new fields in metaphysics

src/Entities/CollectionFeaturedArtwork.ts
```diff
@@ -0,0 +1,52 @@
+import { Field, ID, Int, ObjectType } from "type-graphql"
+import { Column, Entity, ObjectID, ObjectIdColumn } from "typeorm"
+
+@ObjectType()
+@Entity()
+export class CollectionFeaturedArtworkImage {
```
l2succes: I think we would just name this model `ArtworkImage` as we'll end having to maintain two models that have essentially the same properties

src/utils/saveFeaturedArtworks.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * For a given collection, attach top 3 artworks ranked by merchandisability
+ */
+export const attachFeaturedArtworks = async collection => {
+  try {
+    const featuredArtworks = await getFeaturedArtworks(
+      `{
+        marketingCollection(slug: "${collection.slug}") {
```
l2succes: I don't think we need to do something about it now but here's my concern with the current request path for resolving this query

1. KAWS sends the initial request to Metaphysics 
2. Metaphysics uses schema stitching to resolve `marketingCollection` and makes a request to KAWS
3. Metaphysics makes a request to gravity to resolve the search query for artworks
4. KAWS gets the response back

That's 4 roundtrips. We could cut that to one by using the following query instead
```graphql
{
 filterArtworks(filters...) {
 }
}
```

joeyAghion:
I shared some earlier concerns about this design; specifically, that it duplicates artwork data into a non-authoritative system (with the additional complexity, synchronization, and maintenance that comes with that).

I understand that alternatives have been discussed, but I'm unclear what problem this is still attempting to work around. I'm pretty sure the API can serve the data necessary for these featured artworks. If the designs call for a large number of these new requests to happen and would tolerate stale data, we should explore custom (long-TTL) caching of the results in Metaphysics or downstream. If there's a technical obstacle to that, it's important for us to understand it because otherwise this type of architecture headache will recur elsewhere.

eessex:
@joeyAghion as noted in the PR description we are going to continue exploring the strategy of caching this data in metaphysics separately, but decided to get this in place now to unblock the upcoming test in the meantime. 

@l2succes this PR has been updated based on your comments and is ready for another round of review.

eessex:
Put a hold on this strategy in favor of https://github.com/artsy/metaphysics/pull/1575

Will try implementing the feature using metaphysics and monitor throughput on the gravity filter_artworks endpoint and collections pages to see if this is sufficient.  If not, we can return to the strategy of saving in Kaws.

PR #79, opened 2019-03-12
Use slugify to sanitize slug

eessex:
Removes manual regex in favor of `slugify` package to sanitize slugs when importing.


xtina-starr:
Looks good!

PR #80, opened 2019-03-14
Update Hokusai configuration

eessex:
These changes have not yet been applied to production (though they are active in staging now). This PR addresses all items flagged in https://artsyproduct.atlassian.net/browse/PLATFORM-1151, also tracked via https://artsyproduct.atlassian.net/browse/GROW-1156

- Use kubectl version 1.10.7
- Pin node runtime version and use Alpine build
- Set up `deploy` user
- Use Dumb-Init as `ENTRYPOINT`
- Use npm rather than script to install yarn
- Run `yarn cache clean` inline with `yarn install`
- Update hokusai required version
- Add mongo to hokusai dev build
- Use autoscaling for replicas
- Use `ClusterFirst` for dns
- Update heath check endpoint delay/period/timeout seconds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,10 +28,13 @@ jobs:
       - checkout
       - run:
           name: Configure hokusai
-          command: hokusai configure --kubectl-version 1.6.3 --s3-bucket artsy-citadel --s3-key k8s/config --platform linux
+          command:
+            hokusai configure --kubectl-version 1.10.7 --s3-bucket artsy-citadel
+            --s3-key k8s/config --platform linux
       - run:
           name: Update staging branch
-          command: git push git@github.com:artsy/kaws.git $CIRCLE_SHA1:staging --force
+          command:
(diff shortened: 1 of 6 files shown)


eessex:
These changes have been successfully applied to staging, I'll wait for any comments here before applying to production.

eessex:
Hokusai changes have been applied to production .

.circleci/config.yml
```diff
@@ -28,10 +28,13 @@ jobs:
       - checkout
       - run:
           name: Configure hokusai
-          command: hokusai configure --kubectl-version 1.6.3 --s3-bucket artsy-citadel --s3-key k8s/config --platform linux
+          command:
+            hokusai configure --kubectl-version 1.10.7 --s3-bucket artsy-citadel
+            --s3-key k8s/config --platform linux
```
joeyAghion: Is this line-break intentional and valid, or auto-formatting gone wrong? There's a similar one below.

joeyAghion: Checked this out and loaded it successfully so it seems valid.

PR #85, opened 2019-03-28
Automate pulling CSV using Google sheets API

l2succes:
https://artsyproduct.atlassian.net/browse/GROW-1144

We make use of the Google Sheets API to get update KAWS. The goal is enable admins to update and ingest data without engineering intervention. This is only one part of enabling this, the google sheet script needs this endpoint to post the current googlesheet id the user wants to update.

In terms of security, this script uses a JWT which only has permissions to read a particular google sheet from artsy's google account.

1. Adds express route to handle import request from Google Sheet script
2. Extract data sanitization helper method to its own module for reuse
3. Fixes `hokusai dev start` and our docker compose config

### DIFF
--- a/src/utils/sanitizeRow.ts
+++ b/src/utils/sanitizeRow.ts
@@ -1,0 +1,45 @@
+import slugify from "slugify"
+import { Collection } from "../Entities"
+
+export const sanitizeRow = ({
+  title,
+  slug,
+  category,
+  description,
+  headerImage,
+  credit,
+  artist_ids,
+  gene_ids,
+  tag_id,
+  keyword,
+  price_guidance,
+  show_on_editorial,
+  is_featured_artist_content,
+}) => {
+  return {
+    title,
+    slug: sanitizeSlug(slug),
+    category,
+    description,
+    headerImage,
+    credit,
+    price_guidance: price_guidance ? Number(price_guidance) : null,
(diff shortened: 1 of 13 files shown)


src/utils/sanitizeRow.ts
```diff
@@ -0,0 +1,45 @@
...
+    category,
+    description,
+    headerImage,
+    credit,
+    price_guidance: price_guidance ? Number(price_guidance) : null,
+    show_on_editorial: Boolean(show_on_editorial),
+    is_featured_artist_content: Boolean(is_featured_artist_content),
+    query: (artist_ids || gene_ids || tag_id || keyword) && {
+      artist_ids: artist_ids ? artist_ids.split(",").map(a => a.trim()) : [],
+      gene_ids: gene_ids ? gene_ids.split(",").map(a => a.trim()) : [],
```
anandaroop: Do admins always strictly enter these as `id1,id2`, as opposed to `id1, id2` (i.e. does this need to be more lenient and split on comma + whitespace?)

l2succes:
Awesome! thanks @anandaroop

src/utils/sanitizeRow.ts
l2succes: that's what the `.trim()` method is for, it removes whitespace from both sides of the string after we split on `,`.

PR #88, opened 2019-04-03
Calculate price guidance

xtina-starr:
Addresses: [GROW-1173](https://artsyproduct.atlassian.net/browse/GROW-1173)

This adds a util to calculate the base price for works in a collection based on the average of the lowest three prices.

### DIFF
--- a/src/utils/__tests__/getPriceGuidance.test.ts
+++ b/src/utils/__tests__/getPriceGuidance.test.ts
@@ -13,15 +13,12 @@ describe("#getPriceGuidance", () => {
           hits: [
             {
               price: "$295",
-              is_price_range: false,
             },
             {
               price: "$550",
-              is_price_range: false,
             },
             {
               price: "$650",
-              is_price_range: false,
             },
           ],
         },
@@ -33,5 +30,21 @@ describe("#getPriceGuidance", () => {
 
     expect(mockMetaphysics.mock.calls[0][0]).toContain("kaws-companions")
     expect(avgPrice).toEqual(498)
+  })
+
+  it("returns null when the artworks in the collection are not acquireable", async () => {
+    const results = {
+      marketingCollection: {
+        artworks: {
+          hits: [],
+        },
+      },
+    }
+
+    mockMetaphysics.mockResolvedValue(results)
+    const avgPrice = await getBasePrice("josef-albers-never-before")
+
+    expect(mockMetaphysics.mock.calls[0][0]).toContain("kaws-companions")
+    expect(avgPrice).toEqual(null)
   })
 })
--- a/src/utils/__tests__/getPriceGuidance.test.ts
+++ b/src/utils/__tests__/getPriceGuidance.test.ts
@@ -13,13 +13,19 @@ describe("#getPriceGuidance", () => {
         artworks: {
           hits: [
             {
+              price: "$225",
+            },
+            {
+              price: "$225",
+            },
+            {
               price: "$295",
             },
             {
-              price: "$550",
+              price: "$350",
             },
             {
-              price: "$650",
+              price: "$360",
             },
           ],
         },
@@ -30,7 +36,7 @@ describe("#getPriceGuidance", () => {
     const avgPrice = await getPriceGuidance("kaws-companions")
 
     expect(mockMetaphysics.mock.calls[0][0]).toContain("kaws-companions")
-    expect(avgPrice).toEqual(498)
+    expect(avgPrice).toEqual(300)
   })
 
   it("returns null when the artworks in the collection are not acquireable", async () => {
--- a/src/utils/__tests__/getPriceGuidance.test.ts
+++ b/src/utils/__tests__/getPriceGuidance.test.ts
@@ -1,0 +1,37 @@
+import { getBasePrice } from "../getPriceGuidance"
+jest.mock("../../lib/metaphysics", () => ({
+  metaphysics: jest.fn(),
+}))
+
+const mockMetaphysics = require("../../lib/metaphysics").metaphysics
+
+describe("#getPriceGuidance", () => {
+  it("calculates the average of the 3 lowest artworks of a collection", async () => {
+    const results = {
+      marketingCollection: {
+        artworks: {
+          hits: [
+            {
+              price: "$295",
+              is_price_range: false,
+            },
+            {
+              price: "$550",
+              is_price_range: false,
+            },
+            {
+              price: "$650",
+              is_price_range: false,
+            },
+          ],
+        },
+      },
+    }
+
+    mockMetaphysics.mockResolvedValue(results)
+    const avgPrice = await getBasePrice("kaws-companions")
+
+    expect(mockMetaphysics.mock.calls[0][0]).toContain("kaws-companions")
+    expect(avgPrice).toEqual(498)
+  })
+})
--- a/src/utils/__tests__/getPriceGuidance.test.ts
+++ b/src/utils/__tests__/getPriceGuidance.test.ts
@@ -1,4 +1,4 @@
-import { getBasePrice } from "../getPriceGuidance"
+import { getPriceGuidance } from "../getPriceGuidance"
 jest.mock("../../lib/metaphysics", () => ({
   metaphysics: jest.fn(),
 }))
@@ -26,7 +26,7 @@ describe("#getPriceGuidance", () => {
     }
 
     mockMetaphysics.mockResolvedValue(results)
-    const avgPrice = await getBasePrice("kaws-companions")
+    const avgPrice = await getPriceGuidance("kaws-companions")
 
     expect(mockMetaphysics.mock.calls[0][0]).toContain("kaws-companions")
     expect(avgPrice).toEqual(498)
@@ -42,7 +42,7 @@ describe("#getPriceGuidance", () => {
     }
 
     mockMetaphysics.mockResolvedValue(results)
-    const avgPrice = await getBasePrice("josef-albers-never-before")
+    const avgPrice = await getPriceGuidance("josef-albers-never-before")
 
     expect(mockMetaphysics.mock.calls[0][0]).toContain("kaws-companions")
     expect(avgPrice).toEqual(null)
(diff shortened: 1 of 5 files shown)


src/utils/getPriceGuidance.ts
```diff
@@ -1,10 +1,13 @@
 import { reduce } from "lodash"
-const { metaphysics } = require("../lib/metaphysics")
+import metaphysics from "../lib/metaphysics"
+// I notice when I don't require the metaphysics module
+// like below the test git satfails
+// const metaphysics = require("../lib/metaphysics").metaphysics
```
xtina-starr: @eessex do you know why the tests require the module to be loaded in using the `require` method?

eessex: Hm I haven't experienced this issue, just checked this pr and it looks like we were able to use the import like above-- https://github.com/artsy/kaws/pull/78/files

src/utils/__tests__/getPriceGuidance.test.ts
```diff
@@ -0,0 +1,50 @@
+import { getPriceGuidance } from "../getPriceGuidance"
+jest.mock("../../lib/metaphysics", () => ({
+  metaphysics: jest.fn(),
+}))
+
+const mockMetaphysics = require("../../lib/metaphysics").metaphysics
+
```
eessex: Just checked other mockings of this, i think the issue is that you want the default rather than named export: `jest.mock("../../lib/metaphysics", () => ({ default: jest.fn() }))`

xtina-starr: Ahhh, thanks!

src/utils/getPriceGuidance.ts
```diff
@@ -0,0 +1,45 @@
+import { reduce } from "lodash"
+import metaphysics from "../lib/metaphysics"
+const currency = require("currency.js")
+
+const formatCurrency = value => currency(value, { separator: "" }).format()
```
eessex: I'm curious to know what the saved value looks like? From here it appears to be `1000` -- but also curious if the averages do floats `1000.95` etc.

xtina-starr: I'm rounding floats up if necessary on line [44](https://github.com/artsy/kaws/blob/cb48af66868cd40a1af4889e1001da834403daea/src/utils/getPriceGuidance.ts#L44).  So in that case `1000.95` woule be `1001`.

joeyAghion:
Does this account for currencies? How can we average prices otherwise?

In anticipation of our internal guidelines, this new value should probably be stored with a currency code and in minor units (e.g., cents), with any formatting being applied on top of that.

xtina-starr:
@joeyAghion currently to calculate price average/price guidance for the collections we are filtering by [`acuireable`](https://github.com/artsy/gravity/blob/4a7d28b5954e1c52e44bc1522f76ca621b535517/app/models/util/for_sale.rb#L261-L264) which only have price currency as USD. I believe acquireable works only being in USD is subject to change but the data modeling for collections in general could use a bit more consideration. I'd advocate for us to do that keeping in mind price currency and formatting along with the many other factors or current limitations. This change is a quick fix to allow for the `price_guidance` property to be a little bit more scaleable and less manual. 

@eessex I know you have worked on the `price_guidance` in the past. Do you have any additional thoughts?

eessex:
As far as I know the manually entered `price_guidance` is a number (not float) in USD -- but I think @joeyAghion makes a good point that we should make the way data is stored consistent to other places in the APIs where we are saving currency. Doing so may require converting manual `price_guidance` data into a different format, but seems worthwhile for future-proofing.

xtina-starr:
Yeah I'm definitely in favor of future proofing but I'm wondering if that's in the scope of this PR. The concept of price guidance itself isn't full finalized. I still think there is a bit more thought that needs to go into how this is stored and used (or if it's even a permanent property). Also price comes in from metaphysics as a string. I'm assuming it is also represented in gravity as a string so those seem like considerations as well. This is currently being used for the purposes of an a/b test. It is still relatively early so running a migration to rectify the current data might not be a big lift but I'm wondering if this could be held off until the presents, use case, etc is a little more clear.

xtina-starr:
I added a few requirement updates [GROW-1173](https://artsyproduct.atlassian.net/browse/GROW-1173)

junybrum:
@joeyAghion @eessex any additional thoughts given @xtina-starr's previous comments? This price guidance is the last piece we're waiting on before being able to launch the test - we will definitely re-evaluate this approach more holistically when we think about upcoming projects to scale KAWS, but this current solution effectively allows us to save the marketing team a ton of manual work and ensures we can get learnings faster from the test.

src/utils/getPriceGuidance.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const getPriceGuidance = async (slug: string) => {
+  const results: any = await metaphysics(`{
+    marketingCollection(slug: "${slug}") {
+      artworks(
+        size: 5,
+        sort: "prices",
+        price_price: "10-*"
```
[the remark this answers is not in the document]
xtina-starr: Definitely a typo. Thanks!

xtina-starr:
@joeyAghion perhaps we can hop on a call about this Monday. Maybe during your pairing/office hours. 

For more context, I tried adding [`priceCents`](https://github.com/artsy/metaphysics/blob/2bc0d6a3a2f4ff669a4bcf67f20d36079397a3ab/src/schema/artwork/index.ts#L500-L526) to the query. That seems to maps to `price_cents` on the artwork in gravity but it often returns `null`.

The query I'm using: 
```JavaScript
{
  marketingCollection(slug: "kaws-companions") {
    artworks(
      size: 5,
      sort: "prices",
      price_range: "10-*"
    ) { 
      hits {
        price
	priceCents {
	  min
	  max
	  exact
	}
        price_currency
      }
    }
  }
}
```

Response:
```JavaScript
"data": {
    "marketingCollection": {
      "artworks": {
        "hits": [
          {
            "price": "$225",
            "priceCents": null,
            "price_currency": "USD"
          },
          {
            "price": "$225",
            "priceCents": null,
            "price_currency": "USD"
          },
          {
            "price": "$295",
            "priceCents": null,
            "price_currency": "USD"
          },
          {
            "price": "$350",
            "priceCents": null,
            "price_currency": "USD"
          },
          {
            "price": "$360",
            "priceCents": null,
            "price_currency": "USD"
          }
        ]
      }
    }
  }
```

mzikherman:
The issue with that field not being present is that it's only included in `public` JSON: https://github.com/artsy/gravity/blob/2500434d68075cc2aa77e0845bccfd8aa9a15ca4/app/models/domain/artwork.rb#L1001

whereas the filter artworks endpoint (and most endpoints which return lists), returns `short` JSON: https://github.com/artsy/gravity/blob/2500434d68075cc2aa77e0845bccfd8aa9a15ca4/app/api/v1/filter_endpoint.rb#L69

It turns out I added that `price_cents` field a few months ago: https://github.com/artsy/gravity/pull/11943 , and that was actually the first time a raw cents field for an artwork was exposed via the API(!). Everything before that was just working on the friendlier display_price/availability/sale_message fields.

This all being said, I would be into making `price_cents` accessible in the short JSON (so @xtina-starr your query above should then work). If we want to encourage and move to a model of serving minor units/cents and currency from backends, then including such critical data in an object's short JSON (where the now-anti-pattern 'friendly' strings like `display_price` and `sale_message` already exist), is in the spirit of that IMO.

joeyAghion:
Thanks @mzikherman! I wouldn't have figured that out. 😅

xtina-starr:
We talked about this at standup yesterday. It seems like we decided to go ahead with calculating the price guidance based on `priceCents` but follow up on adding the currency and priceGuidanceInCents properties in another PR in order to get the test out.

src/utils/getPriceGuidance.ts
```diff
@@ -0,0 +1,66 @@
...
+        ) / collectionPricesInDollars.length
+    }
+  } catch (error) {
+    throw error
+  }
+
+  return hasNoBasePrice ? avgPrice : Math.ceil(avgPrice / 100) * 100
+}
+
+// getPriceGuidance("kaws-bearbrick")
```
eessex: extra comment

src/utils/getPriceGuidance.ts
```diff
@@ -0,0 +1,66 @@
...
+  const results: any = await metaphysics(`{
+    marketingCollection(slug: "${slug}") {
+      artworks(
+        size: 5,
+        sort: "prices",
+        price_range: "10-*"
+      ) { 
+        hits {
+          price
+          priceCents {
```
l2succes: @xtina-starr Not to add more to this convo but shouldn't we also be getting `price_currency` so we can do the conversion to `USD` if needed?

xtina-starr: I was thinking that the price_cents was already converted in [gravity](https://github.com/artsy/gravity/blob/4a7d28b5954e1c52e44bc1522f76ca621b535517/app/models/util/for_sale.rb#L202-L213)

mzikherman: That doesn't do any exchange rate conversion, that just handles 'dollars to cents' in terms of, knowing to convert the inputted price (in 'dollars' or major units) to the minor unit, usually by multiplying by either 100, or 1 (Japanese yen, etc.).

I _think_ that this PR is overall still not really fully currency aware. There's a hardcoded `/ 100` in there. Right now this works since all `acquireable` works are in USD, but this has the potential to be averaging works that are priced in different currencies. The `price_cents` field in Gravity is not in USD, but in whatever the local currency selected.

For this to be able to fully calculate a base USD price by averaging artworks in the collection, it would need to be able to properly convert an artwork priced in JPY, to an equivalent USD (using a current exchange rate). I don't think we actually do this anywhere besides Looker, and Gravity, currently! So there's no existing pattern for that (but we can add exchange rates to Metaphysics) to make those available. Worth keeping in mind for the future (and maybe adding a comment/documenting somewhere)?

joeyAghion:
As discussed on this PR and more recently in slack and face-to-face, the time has come when acquirable works can be in non-USD currencies. That means a currency code should be stored with any fields based on those core price values, and the currency taken into account in any conversions (e.g., minor to major units) or calculations.

This approach was described as a quick-fix for purposes of an A/B test at the time, and I'm not sure what the outcome of that or future plans are. From our discussion it sounds like some notion of a "normalized" price field might be useful for the API to pre-compute. If that's the case, we should discuss the business rules, especially the best as-of date for converting a non-USD value to USD.

PR #89, opened 2019-04-04
Use hokusai circle-context

eessex:
Use circle context `hokusai` for aws key pairs.

Related conversation: https://artsy.slack.com/archives/CA8SANW3W/p1554396462264300

Screen Shot 2019-04-04 at 3 21 54 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,7 @@ workflows:
   default:
     jobs:
       - test:
+          context: hokusai
           filters:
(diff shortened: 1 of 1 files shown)


eessex:
I removed the aws keys from kaws on Circle, and it seems to run the test suite without issues. Going to self-merge to ensure that other build steps work as expected.

izakp:
👍 looks like build and deploy steps passed as well

PR #91, opened 2019-04-08
@mzikherman => [Hokusai] Add context tag to every step for Circle

mzikherman:
h/t @izakp

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,7 @@ workflows:
               ignore:
                 - staging
       - push:
+          context: hokusai
           filters:
             branches:
               only:
@@ -80,12 +81,14 @@ workflows:
           requires:
             - test
       - deploy-staging:
+          context: hokusai
           filters:
             branches:
               only: master
           requires:
             - push
       - deploy-production:
+          context: hokusai
           filters:
             branches:
               only: release


.circleci/config.yml
```diff
@@ -72,6 +72,7 @@ workflows:
               ignore:
                 - staging
       - push:
+          context: hokusai
```
eessex: In the first implementation I had assumed that the context would be inherited by the requirement of the `push` step.  Is it possible that the production deploy failed because it does not require `push`?  In that case we would only need to add it to line 91.

eessex: Actually I see now that `deploy-production` also requires push -- still confused at why the original staging deploy was able to complete without the context.

izakp: @eessex yeah getting a bit lost now - I didn't know the `requires` directive would mean the context is available to other dependent jobs... But the error I see in the Push step appears to be `ERROR: An error occurred (InvalidClientTokenId) when calling the GetCallerIdentity operation: The security token included in the request is invalid.` Which was due to an invalid secret key in the Hokusai context.

> still confused at why the original staging deploy was able to complete without the context.

Yes so https://circleci.com/workflow-run/8324bd8f-48e3-4500-a2d4-e85168cd9503 succeeded... was the `AWS_SECRET_ACCESS_KEY` changed in https://circleci.com/gh/organizations/artsy/settings#contexts/5161a9b5-0695-4683-accf-4812017ad3f3 after this?

izakp: I just corrected the `AWS_SECRET_ACCESS_KEY` and it works so maybe this PR was just unnecessary

eessex: Yes, I had removed the kaws keys before merging to explicitly test that if it deployed correctly

PR #93, opened 2019-04-08
Use inherited circle context

eessex:
I believe that the issues with keys failing on deploy was due to an incorrect key in the context. This removes changes from https://github.com/artsy/kaws/pull/91. The hokusai context should be inherited from the `test` job, which is required for all steps.


eessex:
Going to go ahead and merge this to confirm all works as expected.

PR #96, opened 2019-04-19
Update priceGuidance script to still work with less data

l2succes:
We realized we were being too strict with the price guidance. Namely
- still compute a value if one of the 5 price values is `null`
- still computer a value if we have less than 5 seeds


xtina-starr:
LGTM!

PR #97, opened 2019-05-23
[Related Collections Footer] filter collections by category

xtina-starr:
Addresses: [GROW-1272](https://artsyproduct.atlassian.net/browse/GROW-1272)

While working on [Related Collections Footer Rail](https://artsyproduct.atlassian.net/browse/GROW-1228), I noticed that we don't yet have the ability to pass `category` as an argument to `marketingCollections` and filter by a specific category. This PR adds that capability.

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -81,6 +81,7 @@ scalar DateTime
 
 type Query {
   collections(
+    category: String
     randomizationSeed: String
     size: Int
     isFeaturedArtistContent: Boolean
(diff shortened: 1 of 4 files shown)


_schema.graphql
```diff
@@ -81,6 +81,7 @@ scalar DateTime
 
 type Query {
   collections(
+    category: String
```
l2succes: I was thinking an enum could make sense here but on second thought we don't know what categories are going to look like in the near future.

PR #99, opened 2019-06-05
adds relatedCollections field resolver

xtina-starr:
Addresses: [GROW-1228](https://artsyproduct.atlassian.net/browse/GROW-1228)

This adds `relatedCollections` to be used in the RelatedCollectionsRail as a top level property via a [`FieldResolver`](https://github.com/19majkel94/type-graphql/blob/master/docs/resolvers.md#field-resolvers).

### DIFF
--- a/src/Resolvers/Collections.ts
+++ b/src/Resolvers/Collections.ts
@@ -1,5 +1,5 @@
-import { isEmpty } from "lodash"
-import { Arg, Int, Query, Resolver } from "type-graphql"
+import { isEmpty, reject } from "lodash"
+import { Arg, FieldResolver, Int, Query, Resolver, Root } from "type-graphql"
 import { getMongoRepository } from "typeorm"
 import { Collection } from "../Entities/Collection"
 import { CollectionCategory } from "../Entities/CollectionCategory"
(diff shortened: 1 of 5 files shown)


src/Resolvers/Collections.ts
```diff
@@ -75,6 +75,36 @@ export class CollectionsResolver {
...
+      })
+
+      const relatedArtistCollections = reject(artistResults, {
+        id: collection.id,
+      })
+
+      if (relatedArtistCollections.length > 4) {
+        return relatedArtistCollections
+      }
+    }
```
l2succes: I think you should have a fallback return here. like `return []` otherwise if there are less than 4 related artist collections, it'll continue and fetch related category collections

xtina-starr: That's what we would want based on the requirements, right? If the collection is associated with an artist and that artist has at least 5 collections display those otherwise fallback to display collections with the same category [GROW-1228](https://artsyproduct.atlassian.net/browse/GROW-1228).

src/Resolvers/Collections.ts
```diff
@@ -75,6 +75,39 @@ export class CollectionsResolver {
...
+
+    const relatedCategoryResults = await this.repository.find({
+      where: {
+        category: { $in: [collection.category] },
+        show_on_editorial: true,
```
xtina-starr: I also added a filter for `show_on_editorial` which is also mentioned in the requirements.

src/Resolvers/Collections.ts
l2succes: Yup, you're right. looks good then

PR #100, opened 2019-06-07
update ci config

xtina-starr:
This updates the CI config "Update staging branch" command to explicitly specify a branch rather than a tag and resolves `error: unable to push to unqualified destination: staging` in https://circleci.com/gh/artsy/kaws/436

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,8 @@ jobs:
       - run:
           name: Update staging branch
           command:
-            git push git@github.com:artsy/kaws.git $CIRCLE_SHA1:staging --force
+            git push git@github.com:artsy/kaws.git
+            $CIRCLE_SHA1:refs/heads/staging --force
       - run:
           name: Deploy staging assets
           command: hokusai staging deploy $CIRCLE_SHA1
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,7 +38,8 @@ jobs:
       - run:
           name: Update staging branch
           command:
-            git push git@github.com:artsy/kaws.git $CIRCLE_SHA1:staging --force
+            git push git@github.com:artsy/kaws.git
+            $CIRCLE_SHA1:refs/heads/staging --force
```
joeyAghion: Is the linebreak intentional? Looks broken to me.

xtina-starr: Not intentional--prettier maybe? I'm also seeing the same on line [32 & 33](https://github.com/artsy/kaws/blob/2076f9916b562d1bb127b3a4360b1ec436d4c96f/.circleci/config.yml#L32-L33)

xtina-starr:
@joeyAghion do you have any more feedback for this?

eessex:
This looks good to me as a quick fix if you are blocked, but this setup can also be simplified by using the  [Orb](https://github.com/artsy/orbs/blob/master/src/hokusai/hokusai.yml#L62) for our deploys -- would be happy to pair with you on configuring this, there's an example PR here: https://github.com/artsy/positron/pull/2014

joeyAghion:
Sorry for the delay while I was out!

PR #102, opened 2019-06-13
Release

eessex:
Related collections, CI config, named ports

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -20,6 +20,8 @@ spec:
       containers:
         - name: kaws-web
           env:
+            - name: PORT
+              value: "4000"
             - name: DD_TRACE_AGENT_HOSTNAME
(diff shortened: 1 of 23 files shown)


hokusai/production.yml
```diff
@@ -20,6 +20,8 @@ spec:
       containers:
         - name: kaws-web
           env:
+            - name: PORT
+              value: "4000"
```
l2succes: @eessex Should the port number be in quotes here? if yes, how come?

eessex: Yes -- we actually encountered this as a breaking change in Positron, and further one that `hokusai staging update --dry-run` doesn't catch. All values declared in the `env` block must be strings.

l2succes: Got it! Thanks

PR #107, opened 2019-06-20
Minor updates to dev setup

anandaroop:
I just got my copy of kaws bootstrapped locally, so I wanted to contribute back a few minor updates.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,3 +14,8 @@ METAPHYSICS_URL=
 
 # Database auth
 MONGOHQ_URL=mongodb://localhost:27017/kaws
+
+# Google Sheets credentials
+GOOGLE_CLIENT_ID=REPLACE
+GOOGLE_CLIENT_SECRET=REPLACE
+
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ collections you ask? A Collection is a prefiltered version of Artsy's
 
 [![CircleCI](https://circleci.com/gh/artsy/kaws.svg?style=svg)](https://circleci.com/gh/artsy/kaws)
 
-- **State:** in development
+- **State:** production
 - **Production:** [http://kaws.artsy.net](http://kaws.artsy.net/playground) |
   [k8s](https://kubernetes.artsy.net/#!/deployment/default/kaws-web?namespace=default)
 - **Staging:**
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ yarn dev
 ```
 
 Then open http://localhost:4000/playground and you should see Apollo's GraphQL
-Playground. Try running the GraphQL query below, and if you see
-`"KAWS, Companions"` and `"Pablo Picasso, Lithographs"` you are good to go!
+Playground. Try running the GraphQL query below, and if you see real collections
+data coming back from the database you are good to go!
 
 ```graphql
 query {
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ Usually you do not have to change them, but if you need, update the variables to
 match your local development setup.
 
 Once this is done, the next step would be to load test data. The collection data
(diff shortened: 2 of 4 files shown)


.env.example
```diff
@@ -13,4 +13,9 @@ SAILTHRU_SECRET=
 METAPHYSICS_URL=
 
 # Database auth
-MONGOHQ_URL=mongodb://localhost:27017/kaws
\ No newline at end of file
+MONGOHQ_URL=mongodb://localhost:27017/kaws
+
+# Google Sheets credentials
+GOOGLE_CLIENT_ID=REPLACE
+GOOGLE_CLIENT_SECRET=REPLACE
```
anandaroop: The app will no longer boot unless these are present.

README.md
```diff
@@ -7,7 +7,7 @@ collections you ask? A Collection is a prefiltered version of Artsy's
 
 [![CircleCI](https://circleci.com/gh/artsy/kaws.svg?style=svg)](https://circleci.com/gh/artsy/kaws)
 
-- **State:** in development
+- **State:** production
```
anandaroop: It's time 😉

README.md
```diff
@@ -62,8 +63,8 @@ yarn dev
 ```
 
 Then open http://localhost:4000/playground and you should see Apollo's GraphQL
-Playground. Try running the GraphQL query below, and if you see
-`"KAWS, Companions"` and `"Pablo Picasso, Lithographs"` you are good to go!
+Playground. Try running the GraphQL query below, and if you see real collections
+data coming back from the database you are good to go!
```
anandaroop: These examples are no longer obvious in the output so I made this more generic.

scripts/bootstrapOrUpdate.ts
```diff
@@ -14,7 +14,7 @@ const csvFile = process.argv[2]
 export async function bootstrapOrUpdate(path: string) {
   const data = await convertCSVToJSON(path)
   try {
-    updateDatabase(data)
+    await updateDatabase(data)
```
anandaroop: I found that `yarn update-database ./collections.csv` was actually just falling silently.

@mbilokonsky helped me see that we were blowing past this promise here. Not sure if this worked recently for others, maybe it's just been a while since anyone has tried this?

src/utils/__tests__/indexForSearch.test.ts
```diff
@@ -45,6 +46,12 @@ describe("indexForSearch", () => {
         },
       })
     )
+    originalConsoleLog = console.log
+    console.log = jest.fn()
+  })
+
+  afterEach(() => {
+    console.log = originalConsoleLog
```
anandaroop: The one opinionated change I made was to keep the test runs free of console output. If this output was in fact useful, I'm happy to revert this.

---
#### Before
---

before

---
#### After
---

after

README.md
```diff
@@ -46,7 +46,8 @@ Usually you do not have to change them, but if you need, update the variables to
 match your local development setup.
 
 Once this is done, the next step would be to load test data. The collection data
-is maintained by Artsy's marketing team [in Google Drive as a spreadsheet](https://docs.google.com/spreadsheets/d/1K-FBuIQYiU75ETBEgU0YuexznElKCLi5Tr_P2bqkFZw/edit#gid=23745674)
+is maintained by Artsy's marketing team
+[in Google Drive as a spreadsheet](https://docs.google.com/spreadsheets/d/1K-FBuIQYiU75ETBEgU0YuexznElKCLi5Tr_P2bqkFZw/edit#gid=23745674)
```
anandaroop: Unless noted otherwise, most of these diffs are Prettier whitespace changes.

.env.example
mbilokonsky: where can these values be found?

anandaroop: `hokusai staging env get` would get you the current staging config, including these env vars.

scripts/bootstrapOrUpdate.ts
l2succes: Nice catch

PR #108, opened 2019-06-25
added linked_collections as a property of the collection model

mbilokonsky:
Our input spreadsheet has had three new columns appended to the end - "artist-series", "featured-collections" and "other-collections". A collection becomes a "hub" if it has any values in any of these columns.

The columns take comma-delimited lists of collection slugs. Any populated column is split on commas and each resulting string is assumed to be a valid reference to a collection. The database right now treats these as arrays of strings, rather than as some sort of ORM-style relational link. (I need to better understand our node ORM if we want to do that, but if this works I'm happy to leave it here, up to y'all)

Anyway, these slugs are then added to a label using the new `Subcollection` datatype `{ name: string, members: string[] }` and then get added to our collection models as appropriate.

Previous tests still pass, which means that rolling this out to prod won't break anything if it gets input from an older version of the spreadsheet, but they do add an empty `subcollections` array to the hub. I added a new test, rather than appending to the old one, for the CSV parsing - just to demonstrate a variety of approaches working.


mbilokonsky:
One thing to note is that while I have tested this against sample CSV downloaded from our google sheet I was not able to get the permissions to work to have the sheet submit its data to the `/upload` endpoint while running on localhost and exposing my system through `ngrok`. Either the permissions were wrong because I'm working against a dummy spreadsheet or there's something weird in passing them through ngrok, but either way this needs to be tested in staging against the real database.

scripts/bootstrapOrUpdate.ts
```diff
@@ -17,6 +17,7 @@ export async function bootstrapOrUpdate(path: string) {
     await updateDatabase(data)
     process.exit(0)
   } catch (e) {
+    console.log(e.message)
```
xtina-starr: Is this a left behind `console.log`?

mbilokonsky: nah that's there on purpose, this is run via various CLI tasks and I wanted to leave a trace of what went wrong if it fails. Logging that output out feels appropriate in this context?

xtina-starr: Makes sense.

src/Entities/Collection.ts
```diff
@@ -1,6 +1,7 @@
 import { Field, ID, ObjectType } from "type-graphql"
 import { Column, Entity, Index, ObjectID, ObjectIdColumn } from "typeorm"
 import { CollectionQuery } from "./CollectionQuery"
+import { Subcollection } from "./Subcollection"
```
l2succes: I'm struggling with the naming here a bit because `Subcollection` is not really a collection but a group of collections. I know most word synonymous with `group` in our namespace are already taken. Maybe `CollectionGroup` makes more sense here?

mbilokonsky: Yeah good call, let me take another swipe at this. I almost wanna call it `NamedCollectionGroup` since that's what it really is?

mbilokonsky: Are we okay with `collection.subcollections` as the property name on the hub?

_schema.graphql
```diff
@@ -44,6 +47,11 @@ type CollectionCategory {
   collections: [Collection!]!
 }
 
+type CollectionGroup {
+  name: String!
```
l2succes: could we add a `type` property which would be `artist_series`, `featured` or `other`?
reasoning is, if we decide to change the ordering of the rails in the UI, we can’t rely on the name of the group as its customizable

PR #111, opened 2019-07-02
Delete .npmignore

eessex:
Removing because we are using Yarn


l2succes:
Makes sense since we don't publish kaws but `.npmignore` still has a use even with yarn for published npm packages.

PR #121, opened 2019-07-16
added hard-coded collection hub slugs; added graphQL integration test…

mbilokonsky:
Implements [GROW-1390](https://artsyproduct.atlassian.net/browse/GROW-1390). This adds support for hard-coded CollectionHubs. If we want to make these dynamic in the future we can just swap out the resolver method (currently it's drawing from a hard-coded array).

The cool thing is that this PR also adds support for GraphQL-aware integration tests. We write to the DB, submit a graphQL query and assert that it returns what we expect.

Diff Analysis: 
```
{
  "total_files_changed": 4,
  "test_files_changed": 2,
  "by_type": {
    "ts": 2,
    "snap": 1,
    "graphql": 1
  }
}
```

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -111,4 +111,5 @@ type Query {
   ): [Collection!]!
   categories: [CollectionCategory!]!
   collection(slug: String!): Collection
+  hubCollections: [Collection!]!
 }
--- a/src/Resolvers/Collections.ts
+++ b/src/Resolvers/Collections.ts
@@ -75,6 +75,20 @@ export class CollectionsResolver {
     return await this.repository.findOne({ slug })
   }
 
+  @Query(returns => [Collection])
+  async hubCollections() {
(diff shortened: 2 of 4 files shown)


l2succes:
Looks like the tests are running _before_ the mongoDB instance is done setting up, if you look at the [logs](https://circleci.com/gh/artsy/kaws/523?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). 
we gotta find a way to setup kaws-mongodb before in the hokusai settings

mbilokonsky:
Nah, it was a specific and silly problem. I had the "clear database" and "write new data to database" in the same try/catch. Because "clear database" errors when you do it against an empty database, it was never getting to the part where it writes new data.

l2succes:
Makes sense, thanks for the explanation

PR #123, opened 2019-07-22
Linked collections

l2succes:
The initial schema resolves the `members` array to string ids not Collection object, this is fixed in this PR.

Enables the following query:

```graphql
{
  collection(slug: "street-art-now") {
    id
    slug
    title
    description
    headerImage
    linkedCollections {
      groupType
      members {
         slug
         title
         headerImage
      }
    }
  }
}
```

### DIFF
--- a/src/Entities/Collection.ts
+++ b/src/Entities/Collection.ts
@@ -98,6 +98,6 @@ export class Collection {
   @Field(type => [CollectionGroup], {
     description: "CollectionGroups of this collection",
   })
-  @Column({ default: [] })
+  @Column(type => CollectionGroup)
   linkedCollections: CollectionGroup[]
 }
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -51,10 +51,9 @@ type CollectionCategory {
 }
 
 type CollectionGroup {
-  id: ID!
   groupType: GroupTypes!
   name: String!
-  members: [String!]!
+  members: [Collection!]!
 }
 
 type CollectionQuery {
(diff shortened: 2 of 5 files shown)


mbilokonsky:
ah thanks for that @l2succes I thought that was already wired up, sorry about that!

I saw the broken test, looked like just an out of date snapshot so I patched and pushed up. merge on green as far as I'm concerned, thank you!

src/Entities/Collection.ts
```diff
@@ -98,6 +98,6 @@ export class Collection {
   @Field(type => [CollectionGroup], {
     description: "CollectionGroups of this collection",
   })
-  @Column({ default: [] })
+  @Column(type => CollectionGroup)
```
mbilokonsky: The type specified here says it's a single CollectionGroup, but the type on the data says it's an Array of collection groups - is this a mistake?

l2succes: I was confused about that as well. I double checked the docs and that's the right way to type arrays
https://typeorm.io/#/mongodb/defining-subdocuments-embed-documents

The example they give is 

```typescript
import {Entity, ObjectID, ObjectIdColumn, Column} from "typeorm";

@Entity()
export class User {

    @ObjectIdColumn()
    id: ObjectID;

    @Column()
    firstName: string;

    @Column()
    lastName: string;

    @Column(type => Profile)
    profile: Profile;

    @Column(type => Photo)
    photos: Photo[];

}
```

If you look at `photos`, the type is `Photo` not `Photo[]`

l2succes:
Awesome, thanks for fixing that @mbilokonsky

PR #127, opened 2019-07-30
added cronjobs to prod and staging for updating sailthru

mbilokonsky:
This PR is meant to satisfy [GROW-1381](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=GROW&modal=detail&selectedIssue=GROW-1381&quickFilter=7), which is a request to re-enable automated sailthru updates.

I followed the link the JIRA ticket to the diffusion example, and brought it over while swapping out everything that seemed like it needed to change. I've updated both prod and staging Hokusai config files, and I ran the task locally using staging data - it succeeded. 

I'd appreciate some :eyes: on my .yaml changes, though, this isn't a space I've spent much time exploring. Tagging @izakp on this since he's the Hokusai king!

Diff Analysis: 
```
{
  "total_files_changed": 2,
  "test_files_changed": 0,
  "by_type": {
    "yml": 2
  }
}
```

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -137,6 +137,7 @@ spec:
                 - configMapRef:
                     name: kaws-environment
           restartPolicy: Never
+          backoffLimit: 2
           affinity:
             nodeAffinity:
               preferredDuringSchedulingIgnoredDuringExecution:
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -113,3 +113,37 @@ spec:
     component: web
   sessionAffinity: None
   type: LoadBalancer
+---
+apiVersion: batch/v2alpha1
+kind: CronJob
+metadata:
+  name: kaws-update-sailthu
+spec:
+  schedule: 13 13 * * *
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: kaws-update-sailthru
+              image: 585031190124.dkr.ecr.us-east-1.amazonaws.com/kaws:production
+              args:
+                - "yarn"
+                - "run"
+                - "update-sailthru"
+              imagePullPolicy: Always
+              envFrom:
+                - configMapRef:
+                    name: kaws-environment
+          restartPolicy: Never
+          affinity:
+            nodeAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+                - weight: 1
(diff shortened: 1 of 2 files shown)


mbilokonsky:
I assume once we merge this we just ensure that it ran correctly the next day and if so we're good to deploy it?

hokusai/production.yml
```diff
@@ -113,3 +113,37 @@ spec:
     component: web
   sessionAffinity: None
   type: LoadBalancer
+---
+apiVersion: batch/v2alpha1
+kind: CronJob
+metadata:
+  name: kaws-update-sailthu
+spec:
+  schedule: 13 13 * * *
```
anandaroop: Wondering if we need to give any thought to the time of day this runs. Is this scheduled to run at daily at 13:13 GMT, as written?

hokusai/staging.yml
```diff
@@ -113,3 +113,37 @@ spec:
     component: web
   sessionAffinity: None
   type: LoadBalancer
+---
+apiVersion: batch/v2alpha1
+kind: CronJob
+metadata:
+  name: kaws-update-sailthu
```
anandaroop: General question — do we actually want to run this cron job in _staging_ as well? Do we even have separate production and staging environments in Sailthru? Or would we end up clobbering our single Sailthru instance with our staging Kaws data?

hokusai/production.yml
mbilokonsky: This is ported directly from diffusion but am happy to tweak. I made the sailthru indexing job (separate PR) run at 715am every day.

hokusai/staging.yml
mbilokonsky: Staging and Prod have different sailthru credentials in their environments, so I _think_ this is fine - plus it lets us confirm on staging that it's working.

anandaroop: 👌

izakp:
LGTM, just the same note as in https://github.com/artsy/kaws/pull/129 - be aware k8s Jobs do, by default retry, so if you don't want to do that, set the Job template's `.spec.backoffLimit`

hokusai/production.yml
l2succes: That would mean the job runs daily at 9:13 EST which sounds good 👍 
The actual time doesn't matter much for this job as long as it runs daily.

PR #129, opened 2019-07-30
Add CronJobs for updating elastic search

mbilokonsky:
NOTE: I based this PR on [this one](https://github.com/artsy/kaws/pull/127), both add new cronjobs and it seemed conflicty to do two from scratch. We should merge that one first, then rebase this one if necessary and merge it.

I ran this locally against staging data and it seemed to work without a hitch.

Like before, since this is Hokusai-related I'm tagging @izakp, and @anandaroop I've got you as the assignee but happy to change that?

Diff Analysis: 
```
{
  "total_files_changed": 2,
  "test_files_changed": 0,
  "by_type": {
    "yml": 2
  }
}
```

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -136,7 +136,7 @@ spec:
               envFrom:
                 - configMapRef:
                     name: kaws-environment
-          restartPolicy: Neve
+          restartPolicy: Never
           affinity:
             nodeAffinity:
               preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


mbilokonsky:
@izakp added that `backoffLimit` to dev and prod yaml, both here and in the other PR. Did I put it in the right place? Does that all look good?

izakp:
@mbilokonsky looks good to me.  P.S. you can validate these files with `hokusai staging/production update --dry-run`

PR #137, opened 2019-08-22
GROW-1408 - validate Kaws inputs

mbilokonsky:
This satisfies the requirements in [GROW-1408](https://artsyproduct.atlassian.net/browse/GROW-1408) as I understood them.

The major change is just this: whenever you upload data into KAWS that includes linked collection slugs, those linked collections must also be included in the set of collections your posting.

I'm PRing this right now even though I have three local test failures stemming from my local mongo being hosed. I've tried reinstalling it twice and it still won't start, I think I screwed up permissions somewhere and need to take some time to fix it. But I don't wanna hold up this PR, so assuming that those tests will pass given a mongo collection I _think_ this PR is good to go. 

But @l2succes the ticket did say that we were going to solicit more feedback from the collections team about what else to validate. I think this was the big important one, but is there anything else you'd want me to throw in?

Diff Analysis: 
```
{
  "total_files_changed": 7,
  "test_files_changed": 4,
  "by_type": {
    "ts": 5,
    "csv": 2
  }
}
```


src/utils/processInput.ts
```diff
@@ -1,7 +1,36 @@
...
 
-export const sanitizeRow = ({
+const validate = (input) => {
+  const by_slug = input.reduce((acc, val) => ({ ...acc, [val.slug]: true }), {})
+  const bad_slugs = new Set();
+
+  const process_link = slug_string => slug_string
+    .split(",")
+    .forEach(slug => {
+      if (slug && !by_slug[slug]) {
```
l2succes: 👌🏾

PR #140, opened 2019-09-03
Update hard-coded hubs slugs to correspond to real data

anandaroop:
https://artsyproduct.atlassian.net/browse/GROW-1494

Real Hubs data has been prepared and will be uploaded to kaws shortly. This gets the collections resolver updated to use the new slugs.

#trivial


pepopowitz:
I removed the WIP and I'm going to merge this. This logic doesn't currently work anyway, and merging now will allow us to do a single deploy to get everything we need into staging for the real hubs data.

PR #146, opened 2019-09-06
[GROW-1399] ensure SSL for all requests

xtina-starr:
Addresses: [GROW-1399](https://artsyproduct.atlassian.net/browse/GROW-1399)

This ensures SSL for every request. This redirects to `https` when the request protocol isn't via https://.

### DIFF
--- a/src/lib/__tests__/ensureSSL.test.ts
+++ b/src/lib/__tests__/ensureSSL.test.ts
@@ -1,0 +1,32 @@
+import { ensureSSL } from "../ensureSSL"
+
+describe("Ensure SSL middleware", () => {
+  let req
+  let res
+  let next
+
+  beforeEach(() => {
+    req = { params: {}, logout: jest.fn(), url: "/terms" }
+    res = { redirect: jest.fn() }
+    next = jest.fn()
+  })
+
+  it("redirects normal http requests to https", () => {
+    req.protocol = "http:"
+
+    ensureSSL(req, res, next)
+    console.log("redirect:", res.redirect.mock.calls)
+    expect(res.redirect.mock.calls[0]).toEqual([
+      301,
+      "https://foobar.com/terms",
+    ])
+  })
+
+  it("does not redirect https protocols", () => {
+    req.protocol = "https:"
+
+    ensureSSL(req, res, next)
+    expect(res.redirect).not.toBeCalled()
+    expect(next).toBeCalled()
+  })
+})
--- a/src/lib/__tests__/ensureSSL.test.ts
+++ b/src/lib/__tests__/ensureSSL.test.ts
@@ -15,7 +15,6 @@ describe("Ensure SSL middleware", () => {
     req.protocol = "http:"
 
     ensureSSL(req, res, next)
-    console.log("redirect:", res.redirect.mock.calls)
     expect(res.redirect.mock.calls[0]).toEqual([
       301,
       "https://foobar.com/terms",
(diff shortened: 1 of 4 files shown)


src/lib/ensureSSL.ts
```diff
@@ -0,0 +1,17 @@
+/*
+* Makes sure that any http requests get redirected to https
+*/
+
+import { parse } from "url"
+const { APP_URL } = process.env
+
+export const ensureSSL = (req, res, next) => {
+  const protocol = req.protocol
+  const isSecure = req.secure || protocol === "https:"
```
pepopowitz: Can you give an example of when `req.secure` would be false but `protocol === 'https:'` would be true? (This is most likely a question for my own education than anything else.)

src/lib/ensureSSL.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  if (!isSecure && parse(APP_URL as string).protocol === "https:") {
+    res.redirect(301, APP_URL + req.url)
```
pepopowitz: Would it make sense to construct the new URL by replacing the protocol of the current URL with `https:` instead of looking to a .env variable (which we could potentially forget to set)?

src/lib/ensureSSL.ts
xtina-starr: @pepopowitz in theory, just checking against protocol should be fine. Originally I only had `req.secure` which returns true if the protocol is `https:` but I added `req.protocol` because I thought that made it a little more clear of what was happening in the test.

I can remove the `req.secure` check.

src/lib/ensureSSL.ts
xtina-starr: I think `APP_URL` is something we should always be setting. But the reason I went with this approach was because generally we don't require https for local / dev environments (we also don't have it set up - we get the error `This site can’t provide a secure connection` when trying to navigate to https://localhost:4000/playground). 

This allows us to connect via https in staging and production environments but not require it locally.

src/lib/ensureSSL.ts
pepopowitz: No problem, let's leave it as-is!

src/lib/ensureSSL.ts
pepopowitz: Makes sense - thank you!

src/lib/__tests__/ensureSSL.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    ensureSSL(req, res, next)
+    expect(res.redirect.mock.calls[0]).toEqual([
+      301,
+      "https://foobar.com/terms",
+    ])
+  })
+
+  it("does not redirect https protocols", () => {
+    req.protocol = "https:"
```
izakp: I think the problem is the line `req.protocol = "https:"` 

From the Express docs https://expressjs.com/en/api.html#req.protocol

> When the trust proxy setting does not evaluate to false, this property will use the value of the X-Forwarded-Proto header field if present. This header can be set by the client or by the proxy.

The app should use the `X-Forwarded-Proto` header and so the trust proxy setting should be truthy.

Also this should be checking for `https` _not_ `https:`  (the protocol does not include the `:` delimiter

src/lib/ensureSSL.ts
izakp: This may be the source of the problem - in my experience `req.protocol` evaluates to `http` or `https`, but here with `"https:"` and including the `:` delimiter (which is usually parsed out) this would never evaluate to true

PR #149, opened 2019-09-11
[GROW-1399] Ensure SSL for all request

xtina-starr:
Addresses: [GROW-1399](https://artsyproduct.atlassian.net/browse/GROW-1399)

This is part 2 of https://github.com/artsy/kaws/pull/148 and the [second attempt](https://github.com/artsy/kaws/pull/147) to ensure we make request via SSL.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -2,6 +2,7 @@ NODE_ENV=test
 PORT=4000
 APP_NAME=kaws-test
 DD_APM_ENABLED=false
+APP_URL=https://foobar.com
 
 # Database auth
 MONGOHQ_URL=mongodb://localhost:27017/kaws-test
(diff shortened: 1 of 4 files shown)


.env.test
```diff
@@ -2,6 +2,7 @@ NODE_ENV=test
 PORT=4000
 APP_NAME=kaws-test
 DD_APM_ENABLED=false
+APP_URL=https://foobar.com
```
eessex: Might be clearer to use the real url here, I'm assuming we make no actual requests in tests.

src/index.ts
```diff
@@ -71,6 +72,12 @@ async function bootstrap() {
     // Setup middleware
     app.use(morgan("combined"))
 
+    // Make sure we're using SSL
+    if (process.env.NODE_ENV !== "development") {
```
eessex: We are already importing a few vars from `process.env` at the top, can import `NODE_ENV` there as well

PR #151, opened 2019-09-13
update-site-map job to production yml

mbilokonsky:
This is sort of just a copy/paste really with a few variables changed. I tried to run the `dry-run` command but my local kubernetes/hokusai config wouldn't let me run it and I got pulled into something else before I could debug it, now I just wanna submit this PR so I don't forget.

Diff Analysis: 
```
{
  "total_files_changed": 1,
  "test_files_changed": 0,
  "by_type": {
    "yml": 1
  }
}
```

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -212,3 +212,38 @@ spec:
                         values:
                           - background
       backoffLimit: 2
+---
+apiVersion: batch/v2alpha1
+kind: CronJob
+metadata:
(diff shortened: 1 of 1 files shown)


mbilokonsky:
Hey, @eessex helped me understand that I was running the wrong dry-run command. Here's a screen cap of this configuration validating, so I am going to merge this - just leaving the comment for posterity.

image

PR #153, opened 2019-09-16
@eessex => [Error Handling] Return a 404 in error extensions for non existent collections

mzikherman:
Rather than silently fail to find a particular collection (https://www.artsy.net/collection/blah currently 500's, for instance), we now throw an appropriate error.

This gets the customary `errors` to show up in the response.

We then use some custom error formatting to decorate the error with a 404 if it's a 'not found' error. We should then be able to propagate this thru to Metaphysics and Reaction (and switch those 500's to 404's!).

### DIFF
--- a/src/utils/errorHandling.ts
+++ b/src/utils/errorHandling.ts
@@ -1,0 +1,25 @@
+import { GraphQLError, GraphQLFormattedError } from "graphql"
+import { EntityNotFoundError } from "typeorm/error/EntityNotFoundError"
+
+type WriteablePartial<T> = { -readonly [P in keyof T]+?: T[P] }
+
+export const formatError = (err: GraphQLError) => {
+  const result: WriteablePartial<GraphQLFormattedError> = {
+    message: err.message,
+  }
+
+  if (err.locations) {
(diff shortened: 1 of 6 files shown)


src/utils/errorHandling.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  if (err.path) {
+    result.path = err.path
+  }
+
+  const originalError = err.originalError
+  if (originalError instanceof EntityNotFoundError) {
+    result.extensions = { code: 404 }
+  }
+
```
eessex: Is it possible for `findOneOrFail` to error on cases other than 404, for example malformed slug or some other common Kaws error? Not sure if we are returning all errors here or just 404s.

mzikherman: `findOneOrFail` specifically rejects + throws the `EntityNotFoundError`, which I think is generally safe to equate to a 404. If whatever criteria was being used to perform a find didn't return a result, that should be a 404 even if the slug is malformed.

mzikherman:
Merging!

PR #154, opened 2019-09-20
better parsing of private key environment variable

mbilokonsky:
As far as I can tell we are having an issue where the environment variable representing our google cloud compute access is not being parsed correctly. There is a LOT of ambiguity around whitespace and how we deal with this stuff.

As of this PR, the `GOOGLE_CLIENT_SECRET` environment variable for KAWS must be treated as a single line string. This means *keep the `\n` line breaks in* - when node loads the `.env` file, though, it escapes `\n` to `\\n` because it thinks we want that character.

We don't - but we have to escape it. That's what is happening in `authorize` now. 

This _MAY NOT_ fix our core problem, but until this fix is in any solution is going to be more difficult. I'd love to get this merged ASAP!

Diff Analysis: 
```
{
  "total_files_changed": 1,
  "test_files_changed": 0,
  "by_type": {
    "ts": 1
  }
}
```

### DIFF
--- a/src/utils/gSheetDataFetcher.ts
+++ b/src/utils/gSheetDataFetcher.ts
@@ -1,8 +1,6 @@
 import { google } from "googleapis"
 
-const { GOOGLE_CLIENT_ID, GOOGLE_CLIENT_SECRET } = process.env
-
-if (!GOOGLE_CLIENT_ID || !GOOGLE_CLIENT_SECRET) {
+if (!process.env.GOOGLE_CLIENT_ID || !process.env.GOOGLE_CLIENT_SECRET) {
   console.error(
     "[kaws] Please set GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET in env variables"
   )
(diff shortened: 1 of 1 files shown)


mbilokonsky:
Self-assigning because I'm not sure who else to assign here - I need to deploy this to ensure it solves the problems we're seeing, and then we need to update our environment variables.

@joeyAghion I decided to go with this approach rather than base64 because it introduces one less step and allows us to keep working with these values as they are. If you think it makes sense to do a total overhaul to use base64 can do that too, but this passes my smoke test locally and I just wanna see if it works on staging.

Rotating tokens etc is not going to help us if the issue is that we somehow lost the ability to read a token! :)

src/utils/gSheetDataFetcher.ts
```diff
@@ -13,7 +11,16 @@ export const gSheetDataFetcher = async (
...
   sheetName: string
 ) => {
-  const jwtClient = await authorize()
+  let jwtClient
+  try {
+    jwtClient = await authorize()
+  } catch (e) {
+    console.dir(e.message)
+    throw new Error(
+      "Error generating JWT client for google - ensure your GOOGLE_CLIENT_SECRET is a single string with `\\n` for each linebreak. See console output for full error message."
```
joeyAghion: I'm not familiar with typescript backslash-escaping. Will this print 2 backslashes in the error message or 1?

mbilokonsky:
```
~/Projects/work/kaws [GROW-1540 ✅]
myk@mykbook😙 $ node
> console.log('this is a test of `\\n` how this will log out.')
this is a test of `\n` how this will log out.
```

PR #155, opened 2019-09-20
remove await to prevent timeout, document google credential weirdness

mbilokonsky:
Removes an 'await' that was causing timeouts on long bulk operations. Updates documentation.

Diff Analysis: 
```
{
  "total_files_changed": 2,
  "test_files_changed": 0,
  "by_type": {
    "ts": 1,
    "md": 1
  }
}
```

### DIFF
--- a/src/Routes/GSheetImport.ts
+++ b/src/Routes/GSheetImport.ts
@@ -86,8 +86,14 @@ export const upload = async (
   })
 
   try {
-    await updateDatabase(rows)
-    res.send(200)
+    updateDatabase(rows)
+    res
+      .status(200)
+      .send(
+        `Updating ${
+          rows.length
+        } records in the database, please wait a few minutes!`
+      )
   } catch (e) {
     console.log("There was an error uploading collection data.")
     console.log(e.message)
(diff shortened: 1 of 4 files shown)


src/Routes/GSheetImport.ts
```diff
@@ -86,8 +86,14 @@ export const upload = async (
...
-    await updateDatabase(rows)
-    res.send(200)
+    updateDatabase(rows)
+    res
+      .status(200)
+      .send(
+        `Updating ${
+          rows.length
+        } records in the database, please wait a few minutes!`
+      )
```
joeyAghion: I think it's dangerous [as discussed] to just ignore failures calling to the database. Looking at `updateDatabase`, it seems to just iterate through collections doing an upsert for each, which should _not_ be unreliable. Is it possible the `getPriceGuidance` call is the one that might be buggy or erroring out?

mbilokonsky: I hear you and have added some try/catch and logging around `getPriceGuidance` - if that IS what's causing the issue we'll at least be notified of it in the logs.

joeyAghion:
Assuming this works like javascript, I suspect some of those `console.log`s should actually be `console.error` for visibility.

PR #159, opened 2019-09-26
added some sanity and data health checks

mbilokonsky:
We were having some failures in a cron job as documented in [GROW-1549](https://artsyproduct.atlassian.net/browse/GROW-1549) so this warranted a closer look.

I found a few things - rather than leave details here I'll leave my own PR Review comments outlining the choices I made.

Diff Analysis: 
```
{
  "total_files_changed": 2,
  "test_files_changed": 0,
  "by_type": {
    "ts": 2
  }
}
```

### DIFF
--- a/src/utils/indexForSearch.ts
+++ b/src/utils/indexForSearch.ts
@@ -10,18 +10,36 @@ import { search } from "../lib/search"
  * This indexes an ElasticSearch cluster with collections data.
  */
 
+interface DebugRecord {
+  slug: string
+  reason: string
+  data: string
+}
+
 export const indexForSearch = async () => {
   let connection: Connection | undefined
-
+  const bogus_collections: DebugRecord[] = []
   try {
     const connectionArgs = databaseConfig()
     connection = await createConnection(connectionArgs)
 
     if (connection.isConnected) {
       const repository = getMongoRepository(Collection)
-
       const collections = await repository.find()
       for (const collection of collections) {
+        console.log(`Now processing ${collection.slug}`)
+
+        if (!collection.query) {
+          console.log(
+            `\t[error] ${collection.slug} - no query found, skipping!`
+          )
+          bogus_collections.push({
+            slug: collection.slug,
+            reason: "no query found",
+            data: "",
+          })
+          continue
+        }
         // Schema assumes fields named `name`, `featured_names`, `alternate_names`
         // to be present for text search through those fields.
         // Additionally, `visible_to_public` and `search_boost` are required for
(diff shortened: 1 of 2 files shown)


src/utils/indexForSearch.ts
```diff
@@ -10,18 +10,37 @@ import { search } from "../lib/search"
  * This indexes an ElasticSearch cluster with collections data.
  */
 
+interface DebugRecord {
```
mbilokonsky: This is just a type so that the compiler stops yelling at me - I want to log out the set of all records that had any kind of issue and so were unable to be updated.

src/utils/indexForSearch.ts
```diff
@@ -10,18 +10,36 @@ import { search } from "../lib/search"
...
+  slug: string
+  reason: string
+  data: string
+}
+
 export const indexForSearch = async () => {
   let connection: Connection | undefined
-
+  const bogus_collections: DebugRecord[] = []
```
mbilokonsky: This is where we track failures

src/utils/indexForSearch.ts
```diff
@@ -10,18 +10,36 @@ import { search } from "../lib/search"
...
   try {
     const connectionArgs = databaseConfig()
     connection = await createConnection(connectionArgs)
 
     if (connection.isConnected) {
       const repository = getMongoRepository(Collection)
-
       const collections = await repository.find()
       for (const collection of collections) {
+        console.log(`Now processing ${collection.slug}`)
```
mbilokonsky: Let's log out the slug of the collection we're looking at right away, so if we get a crash we can go find that record.

src/utils/indexForSearch.ts
```diff
@@ -10,18 +10,36 @@ import { search } from "../lib/search"
...
+
+        if (!collection.query) {
+          console.log(
```
mbilokonsky: we have some stuff in the prod database with no queries defined - I think this happened when we deployed stub collection hubs with no membership leading to a null query, which is what has been causing the failure in our cron jobs. So this is the minimum fix to the current issue.

src/utils/indexForSearch.ts
```diff
@@ -53,28 +71,49 @@ export const indexForSearch = async () => {
...
-            `${collection.slug} - error fetching artworks: `,
+            `\t[error] - ${collection.slug} - error fetching artworks: `,
             error.message
           )
+          bogus_collections.push({
+            slug: collection.slug,
+            reason: "error fetching artworks",
+            data: error.message,
+          })
+          continue
```
mbilokonsky: This is the root cause of the issues we've been seeing where artworks with no images show up in prod data - previously, the code had anticipated this error and logged it out _but it never called `continue`_ so the code continued as if image_url had been set.

src/utils/indexForSearch.ts
```diff
@@ -83,5 +122,11 @@ export const indexForSearch = async () => {
     /* tslint:disable:no-unused-expression */
     connection && connection.close()
     /* tslint:enable:no-unused-expression */
+    if (bogus_collections.length > 0) {
```
mbilokonsky: Finally, let's summarize the results!

src/utils/indexForSearch.ts
```diff
@@ -53,28 +71,49 @@ export const indexForSearch = async () => {
...
+              featured_names,
+              description,
+              slug,
+              visible_to_public,
+              search_boost,
+              image_url,
+            },
+          })
+        } catch (e) {
+          console.log(
```
mbilokonsky: Capture database write errors per-record

src/utils/processInput.ts
```diff
@@ -48,9 +48,9 @@ export const sanitizeRow = ({
         collection: other_collections,
       }
     ),
-    query: (artist_ids || gene_ids || tag_id || keyword) && {
-      artist_ids: splitmap(artist_ids),
-      gene_ids: splitmap(gene_ids),
+    query: {
```
mbilokonsky: Previously, this was creating a null `query` property on collections if they don't have at least one of artist_ids, gene_ids, tag_ids or keywords. This was the direct cause of the current problem where collections without queries were breaking ingestion.

src/utils/indexForSearch.ts
pepopowitz: Is this going to be really noisy & cause us to quickly reach a cap on any logging services?

mbilokonsky:
Nah this runs once a day over a few hundred collections. We were already logging one line out, this just adds another for debugging purposes. I think it's fine in a cron job.

PR #163, opened 2019-10-01
Add internalID fields to kaws schema

zephraph:
This addresses PLATFORM-1673

In MPv2 the semantics of handling ids has changed. We've added a field called `internalID` to all services that reference whatever the database id for that system might be. The specifics of that id is considered special to the system providing it. 

This is an additive change to the schema to duplicate the `id` fields into an `internalID` field. 

From graphql-inspector: 

```
Detected the following changes (3) between schemas:

✔ Field internalID was added to object type Collection
✔ Field internalID was added to object type CollectionQuery
✔ Field internalID was added to object type CollectionGroup
```

### DIFF
--- a/src/Entities/CollectionGroup.ts
+++ b/src/Entities/CollectionGroup.ts
@@ -1,4 +1,4 @@
-import { Field, ObjectType, registerEnumType } from "type-graphql"
+import { Field, ID, ObjectType, registerEnumType } from "type-graphql"
 import { Column, Entity, ObjectID, ObjectIdColumn } from "typeorm"
 import { Collection } from "./Collection"
 
@@ -19,6 +19,7 @@ export class CollectionGroup {
   @ObjectIdColumn()
   id: ObjectID
 
+  @Field(type => ID, { nullable: true })
   @ObjectIdColumn({ name: "id" })
   internalID: ObjectID
 
--- a/src/Entities/CollectionGroup.ts
+++ b/src/Entities/CollectionGroup.ts
@@ -19,6 +19,9 @@ export class CollectionGroup {
   @ObjectIdColumn()
   id: ObjectID
 
+  @ObjectIdColumn({ name: "id" })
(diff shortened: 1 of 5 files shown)


src/Entities/CollectionGroup.ts
```diff
@@ -19,6 +19,10 @@ export class CollectionGroup {
   @ObjectIdColumn()
   id: ObjectID
 
+  @Field(type => ID, { nullable: true })
```
alloy: Why is this nullable? (I see that `id` has no specification at all, but also doesn’t show up in the schema at all?

src/Entities/CollectionQuery.ts
```diff
@@ -8,6 +8,10 @@ export class CollectionQuery {
   @ObjectIdColumn()
   id: ObjectID
 
+  @Field(type => ID, { nullable: true })
+  @ObjectIdColumn({ name: "id" })
+  internalID: ObjectID
```
alloy: Same question here. I get that it follows the declaration above, but assuming that these are DB IDs, I don’t see why they can be nullable.

src/Entities/CollectionGroup.ts
zephraph: I'm not sure about the original design intent. @pepopowitz, @mbilokonsky, or @xtina-starr do any of you know?

mbilokonsky: Sorry I purely cargo-culted adding the original ID field when I wrote this code. If we don't need it feel free to rip it out, I was just following what I saw from other files I'm not as familiar with this ORM.

zephraph:
So, what should be the next steps here? Do I remove the nullability of the IDs? Is that something we follow up with?

alloy:
@zephraph Yes, merge now imo.

@mbilokonsky An ID field being nullable is generally not a reflection of actual DB state—where presumably the backing model always has an ID.

PR #166, opened 2019-10-08
fixes some issues with price guidance

mbilokonsky:
This is a doozy! Building on the research that @anandaroop did when he wrote up [GROW-1570](https://artsyproduct.atlassian.net/browse/GROW-1570) and [GROW-1571](https://artsyproduct.atlassian.net/browse/GROW-1571), my initial approach was going to be to decouple price guidance computation from ingestion and migrate it to a background task or cron job.

However, because of the way we're using the spreadsheet for user input we're in a bit of a pickle because uploading a row with an empty cell MAY mean "compute a default value", but it may also "replace the prior override with a computed default value", and sometimes the default value is still `null`.

So the thinking here is: whenever a sheet is uploaded and has a value for `priceGuidance`, great. We set that value and move on with our lives.

However, if any rows in that sheet don't have a priceGuidance specified we write a `null` to the database for that field. We then, once everything has been loaded, do a second pass where we search for all collections with `null` price guidance and we try to update each of them with correct data if we can generate it. 

I *think* the tests cover all the bases. Some takeaways from this: `TypeORM` is the ORM we are using in this repo, and (I think) we use it to generate the GraphQL schemas that we pipe out to Metaphysics. In this case, we've recently deprecated snake_case variable names - so the `Collection` entity in `TypeORM` has a property called `priceGuidance`, but the **backing field in the database** is still called `price_guidance`.

This would be no big deal if we consistently used TypeORM for our database operations, but we don't - our upload behavior is driven directly through the MongoDB driver and sidesteps the ORM entirely. Debugging that was a bit of a pain in the butt! :)

Diff Analysis: 
```
{
  "total_files_changed": 5,
  "test_files_changed": 3,
  "by_type": {
    "ts": 5
  }
}
```

### DIFF
--- a/src/__tests__/price_guidance_integration.test.ts
+++ b/src/__tests__/price_guidance_integration.test.ts
@@ -1,0 +1,112 @@
+jest.mock("../utils/getPriceGuidance")
+import { getPriceGuidance } from "../utils/getPriceGuidance"
+;(getPriceGuidance as any).mockResolvedValue(1000)
+
+import { Connection, createConnection, MongoRepository } from "typeorm"
+import { databaseConfig } from "../config/database"
+import { Collection } from "../Entities"
+import { updateDatabase } from "../utils/updateDatabase"
+
+const hub_slugs = [
+  "contemporary",
+  "post-war",
+  "impressionist-and-modern",
+  "pre-20th-century",
+  "photography",
+  "street-art",
+]
+const PGs = [100, 100, 100, 100, 200, null]
+const generateMockData = () => {
+  return hub_slugs.map((slug, index) => {
+    return {
+      title: "Pablo Picasso: Portraits",
+      slug,
+      category: "Modern",
+      description:
+        '<p>Although Pablo Picasso worked in a variety of different styles and mediums throughout his prolific career, he never stopped creating portraits of those in his inner circle. Even in his most abstract portraits, Picasso captures his sitters’ personalities. For example, his cubist portrait of his friend and art dealer <i><a href="https://www.artsy.net/artwork/pablo-picasso-portrait-of-art-dealer-ambroise-vollard-1867-1939">Ambroise Vollard</a></i> (1910), shows the figure broken down into geometric, angled shapes, although the viewer can still discover that Vollard was a serious and somewhat gruff individual with his downcast eyes and frown. Picasso is also famous for his portraits of his lovers, which often distort their features into playful abstractions. His sultry sleeping portrait of Marie-Thérèse Walter, <i>Le Rêve</i> (1932) is among the world’s most expensive paintings ever sold, reaching $155 million through a private sale in 2013. Regarding his outlook on portraiture, Picasso was often elusive. “When you start with a portrait and search for a pure form, a clear volume, through successive eliminations, you arrive inevitably at the egg,” he once said, “Likewise, starting with the egg and following the same process in reverse, one finishes with the portrait.”</p>',
+      headerImage: "http://files.artsy.net/images/picasso-portraits.png",
+      credit:
+        "<p>&copy; Pablo Picasso / Artist Rights Society (ARS), New York, NY.</p>",
+      query: {
+        artist_ids: ["4d8b928b4eb68a1b2c0001f2"],
+        gene_ids: ["portrait"],
+        tag_id: "",
+        keyword: "",
+      },
+      price_guidance: PGs[index],
+    }
+  })
+}
+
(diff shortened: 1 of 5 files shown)


src/utils/updateDatabase.ts
```diff
@@ -15,23 +15,8 @@ export async function updateDatabase(collections: Collection[]) {
...
-            )
-            console.log(e.message)
-          }
-        }
-        await collection.update({ slug: entry.slug }, entry, { upsert: true })
-        console.log("Successfully updated: ", entry.slug, entry.title)
-      }
-
-      console.log("Successfully updated collections database")
+      await upsert_data(collection, collections)
```
pepopowitz: This is an unfortunate naming similarity. Is there any renaming we could do to make it clear that one is a collection in the mongo sense, and one is a set of collections in the sense of our business model?

src/utils/__tests__/getPriceGuidance.test.ts
```diff
@@ -14,42 +14,46 @@ describe("#getPriceGuidance", () => {
...
-              priceCents: {
-                min: 36000,
-                max: 36000,
-              },
-            },
-          ],
-        },
+          artworks_connection: {
+            edges:
+              [
```
anandaroop: Hmm shouldn't each element in the `edges` array be on object that has a `node` property?

I.e. I'd expect something like [this](https://user-images.githubusercontent.com/140521/66503303-f94ad080-ea94-11e9-87cb-311e819193cc.png)? But then, tests are working, so I'm not quite sure about this 🤔

src/utils/getPriceGuidance.ts
```diff
@@ -33,7 +37,7 @@ export const getPriceGuidance = async (slug: string) => {
     let avgPrice
     let hasNoBasePrice
     const collectionPricesInDollars = flatMap(
-      results.marketingCollection.artworks.hits,
+      results.marketingCollection.artworks.artworks_connection.edges,
       artwork => {
```
anandaroop: Related to previous comment, I would have thought that what you get here is not an artwork but a node:  `{ node: { priceCents: 42} }`, no?

src/utils/updateDatabase.ts
```diff
@@ -46,3 +31,48 @@ export async function updateDatabase(collections: Collection[]) {
...
+      await collection.update({ slug: entry.slug }, entry, { upsert: true })
+    }
+  } catch (e) {
+    console.log("Error upserting data!")
+    console.log(e.message)
+    throw e
+  }
+}
+
+const update_price_guidance = async collection => {
```
anandaroop: Any reason these two new functions aren't 🐪 ? (I'd prefer to be consistent in at least the cases where we _do_ have control)

src/utils/updateDatabase.ts
```diff
@@ -46,3 +31,48 @@ export async function updateDatabase(collections: Collection[]) {
     /* tslint:enable:no-unused-expression */
   }
 }
+
+const upsert_data = async (collection, collections) => {
+  try {
+    for (const entry of collections) {
+      // if an entry's `price_guidance` is left null in the dataset, null
+      // it out in the database - after this is finished we will compute price
+      // guidance for all collections that have a null value for it.
```
anandaroop: I think this makes sense. To recap, the price guidance column in the sheet has these semantics:

- a blank always indicates "gimme the computed value"
- a value always indicates an override to be used instead of the computed value

Therefore, on first import:
- blanks result in the computed value
- values result in the override

And on re-import:
- blanks result in a _newly_ computed value, whether the blank was always there or newly blanked out
- values result in the override

If correct, this seems 👍 to me. (Upon re-reading, I think the integration spec confirms exactly this,  lmk if that's not right)

src/__tests__/price_guidance_integration.test.ts
```diff
@@ -0,0 +1,112 @@
+jest.mock("../utils/getPriceGuidance")
+import { getPriceGuidance } from "../utils/getPriceGuidance"
+;(getPriceGuidance as any).mockResolvedValue(1000)
```
anandaroop: Hmm, would adding type info to the function signature of `getPriceGuidance` make this unnecessary?

(Not sure myself, I've def been confused by TS + Jest + mocking before)

pepopowitz:
@mbilokonsky I made you the assignee - feel free to merge when you think this is ready. I do think that you have local changes that you haven't pushed, because I am not seeing the changes we paired on the other day.

PR #168, opened 2019-10-17
Add featured_artist_exclusion_ids column to exclude artists from feat…

pepopowitz:
Collaboration with @xtina-starr 
Addresses https://artsyproduct.atlassian.net/browse/GROW-1596

This PR adds a new field to collections named `featuredArtistExclusionIds`. Due to the very broad coverage of the genes we're using to drive our hubs, some hubs are showing unexpected artists under "Featured Artists". With this field, our team can specify artists to exclude from the featured artists section.

We verified locally that with this change, we could add the field to the ingest spreadsheet, ingest it & get it persisted, then read it in a graphQL query.

### DIFF
--- a/src/Routes/__tests__/GSheetImport.test.ts
+++ b/src/Routes/__tests__/GSheetImport.test.ts
@@ -66,6 +66,7 @@ describe("GSheetImport", () => {
             keyword: "Superman",
           },
           linkedCollections: [],
+          featuredArtistExclusionIds: [],
         },
         {
           title: "Anish Kapoor: Etchings",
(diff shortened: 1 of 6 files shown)


xtina-starr:
@pepopowitz looks like we have a legit failure. I think we needed to update the snapshot test.

pepopowitz:
@xtina-starr I tried to sneak in a description of the field we added to the graphQL schema, and didn't realize it would affect snapshots. Hopefully it's fixed now! 🤞

src/Routes/__tests__/GSheetImport.test.ts
```diff
@@ -119,6 +121,7 @@ describe("GSheetImport", () => {
               groupType: GroupType.OtherCollections,
             },
           ],
+          featuredArtistExclusionIds: ["artist-id-for-pablo-picasso"],
```
anandaroop: Kind of minor, but I think as a matter of communication to future us, this might be better as a BSON-looking id, assuming that's what we will ultimately use at the call site (compare to L106 above).

PR #171, opened 2019-11-01
Sanitize string inputs in ingest. 

pepopowitz:
When we [updated the `typeorm` dependency](https://github.com/artsy/kaws/pull/161), it surfaced the fact that we are storing empty `tag_id` and `keyword` fields as empty strings, instead of `null` or not storing them at all. The `typeorm` update stopped returning these empty strings as `null`, and started returning them as empty strings. This caused all artist and collection pages to break, because we started passing an empty `tag_id` query-string param in requests to Gravity. 

Part 1 of the fix was to correct the data that's already stored. I've done this in staging, and will do it in production shortly.

This PR is part 2 - it prevents _new_ data from getting written with empty strings instead of nulls.

### DIFF
--- a/src/utils/__tests__/processInput.test.ts
+++ b/src/utils/__tests__/processInput.test.ts
@@ -1,0 +1,16 @@
+import { sanitizeString } from "../processInput"
+
+describe("processInput", () => {
+  describe("sanitizeString", () => {
+    it.each`
+      value           | expected
+      ${"some-value"} | ${"some-value"}
+      ${undefined}    | ${null}
+      ${""}           | ${null}
+    `("returns $expected when value is $value", ({ value, expected }) => {
+      const result = sanitizeString(value)
+
+      expect(result).toEqual(expected)
+    })
+  })
+})
--- a/src/utils/processInput.ts
+++ b/src/utils/processInput.ts
@@ -52,14 +52,16 @@ export const sanitizeRow = ({
     query: {
       artist_ids: splitmap(artist_ids),
       gene_ids: splitmap(gene_ids),
-      tag_id,
-      keyword,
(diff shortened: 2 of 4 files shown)


src/utils/processInput.ts
```diff
@@ -52,15 +52,17 @@ export const sanitizeRow = ({
...
+      tag_id: sanitizeString(tag_id),
+      keyword: sanitizeString(keyword),
     },
     featuredArtistExclusionIds: splitmap(featured_artist_exclusion_ids),
   } as Collection
 }
 
 const splitmap = text => (text ? text.split(",").map(a => a.trim()) : [])
 
+export const sanitizeString = (text: string) => (text ? text : null)
```
joeyAghion: Didn't realize TS/JS evaluates empty strings as false!

pepopowitz: There are surprises around every corner with JavaScript.

src/utils/__tests__/processInput.test.ts
```diff
@@ -0,0 +1,16 @@
+import { sanitizeString } from "../processInput"
+
+describe("processInput", () => {
+  describe("sanitizeString", () => {
+    it.each`
```
mbilokonsky: Clever, maybe just a bit too clever for me to follow. This is relying on some of the weird tricks with template strings, how it's the one place you can call a function without parens in JS?

pepopowitz: Correct, [tagged template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). The usage for these tests is documented [in the Jest docs](https://jestjs.io/docs/en/api#2-describeeach-table-name-fn-timeout).

PR #172, opened 2019-11-08
Remove unused Image type

xtina-starr:
We were [asked](https://artsy.slack.com/archives/C2TQ4PT8R/p1573214820148300) to add an `id`  field to the KAWS `Image` type, but since we actually aren't using this field  I thought we could go ahead and remove it.


xtina-starr:
I'm going to go ahead and merge this.

PR #175, opened 2019-11-20
fix price-guidance

mbilokonsky:
An October change to the price guidance logic to add a null pointer check also accidentally flipped the sort order of collection items - we were pulling the five most expensive items, not the five least expensive items.

This PR fixes that change. Once it's been merged to staging we should immediately see the following recover to more normal numbers:

image

### DIFF
--- a/src/utils/__tests__/getPriceGuidance.test.ts
+++ b/src/utils/__tests__/getPriceGuidance.test.ts
@@ -1,4 +1,8 @@
-import { getPriceGuidance, getPriceInDollars } from "../getPriceGuidance"
+import {
+  generateQuery,
+  getPriceGuidance,
+  getPriceInDollars,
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/getPriceGuidance.test.ts
```diff
@@ -10,6 +14,33 @@ describe("#getPriceGuidance", () => {
   afterEach(() => {
     mockMetaphysics.mockClear()
   })
+
+  it("requests the five cheapest items in the given collection that still have a price", () => {
```
pepopowitz: Note that we discussed while pairing that this snapshot test is probably the most effective way for us to verify that we are getting the correct data from metaphysics, considering that we don't really have a good place to write "integration" tests for this import script.

PR #179, opened 2019-12-18
Update renovate assignee

eessex:
Saw @mbilokonsky is listing as a point person, updated to assignee for renovate (which mostly PR's bumps for the Artsy Hokusai Orb).


mbilokonsky:
lgtm #mergeongreen

PR #186, opened 2020-01-13
support markdown descriptions

mbilokonsky:
1. Update schema to add `markdownDescription?` and `hasMarkdownDescription = false`
2. Update package.json to include `turndown`
3. Update processInput.ts to apply markdown transform to description html
4. Update tests and snapshots

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "reflect-metadata": "0.1.12",
     "sailthru-client": "3.0.2",
     "slugify": "1.3.4",
+    "turndown": "^5.0.3",
     "type-graphql": "0.17.5",
     "typeorm": "0.2.20"
   },
--- a/src/Entities/Collection.ts
+++ b/src/Entities/Collection.ts
@@ -34,6 +34,19 @@ export class Collection {
   @Column({ nullable: true })
   description?: string
 
+  @Field({
+    nullable: true,
+    description: "Markdown alternate of description field contents.",
+  })
+  @Column({ nullable: true })
+  descriptionMarkdown?: string
+
+  @Field(type => Boolean, {
+    description: "Collection has its description available as markdown",
+  })
+  @Column({ default: false })
+  hasMarkdownDescription: boolean = false
+
   @Field(type => String, {
     nullable: true,
     description: "Background image for the header of the collection page",
--- a/src/utils/__tests__/__snapshots__/createSchema.test.ts.snap
+++ b/src/utils/__tests__/__snapshots__/createSchema.test.ts.snap
@@ -15,6 +15,12 @@ type Collection {
   # Description of the collection which can include links to other collections
   description: String
 
+  # Markdown alternate of description field contents.
+  descriptionMarkdown: String
+
+  # Collection has its description available as markdown
+  hasMarkdownDescription: Boolean!
+
   # Background image for the header of the collection page
   headerImage: String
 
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -79,6 +79,7 @@
     "reflect-metadata": "0.1.12",
     "sailthru-client": "3.0.2",
     "slugify": "1.3.4",
+    "turndown": "^5.0.3",
```
anandaroop: Turndown looks solid 👍

src/Entities/Collection.ts
```diff
@@ -34,6 +34,19 @@ export class Collection {
   @Column({ nullable: true })
   description?: string
 
+  @Field({
+    nullable: true,
+    description: "Markdown alternate of description field contents.",
+  })
+  @Column({ nullable: true })
+  descriptionMarkdown?: string
```
anandaroop: I've gotten this sort of feedback a couple of times when updating MP so I'm going to put this out there for thought —

Since `description` and `descriptionMarkdown` are basically two representations of the same content, I wonder if it wouldn't be more idiomatic to modify the existing field so that…

```graphql
collection(slug: "foo") {
  description # retrieves the html, to preserve compat with existing clients
}

collection(slug: "foo") {
  description(format: "markdown") # retrieves the markdown-flavored version
}
```

(cc @mzikherman who's caught this kind of thing before for me)

If that's tricky for some reason given the `type-graphql` / `typeorm` setup here, then I don't think it's the end of the world.

src/utils/__tests__/__snapshots__/createSchema.test.ts.snap
```diff
@@ -15,6 +15,12 @@ type Collection {
   # Description of the collection which can include links to other collections
   description: String
 
+  # Markdown alternate of description field contents.
+  descriptionMarkdown: String
+
+  # Collection has its description available as markdown
+  hasMarkdownDescription: Boolean!
```
anandaroop: What's the use case for this field?

If we're going with the approach in this PR, wouldn't we have run a migration beforehand so that `hasMarkdownDescription` would always be `true`? (Or at least true when there is any description at all?)

mbilokonsky: good point!

src/Entities/Collection.ts
mbilokonsky: I have the sense we are phasing the HTML out, right? Perhaps we will eventually deprecate `description` or make it point directly to MD stuff?

Not averse to this, but not sure it's necessary?

src/utils/__tests__/__snapshots__/createSchema.test.ts.snap
mbilokonsky: @ashleyjelks what do you think? Should we remove this boolean, or are you using it reaction? We'd need another PR to MP to pull it out which should not be a factor but I'm eyeing it sideways anyway...

ashleyjelks: I think we should keep it bc in Reaction we need to handle the in-betweens states that some collections will be in where it needs to render HTML bc the data re-processing hasn't isn't complete, and also bc some collections simply don't have descriptions so we can default to render nothing when the boolean is false.

src/Entities/Collection.ts
anandaroop: Yeah, I reckon that's fine. We will presumably have the cleansing fire of `ArtworkProductGrouping` at some point in the near future in any case.

src/utils/__tests__/__snapshots__/createSchema.test.ts.snap
anandaroop: I'd be surprised if that text processing took more than a matter of minutes, if not seconds. For that reason this field strikes me as overkill. 

Also I'd say we should be able gracefully handle the absence of any field on the client side, without needing an _extra_ field to tell us whether or not that field is present? Is there something special about the description field in Emission that would require us to have that "default to render nothing" behavior?

ashleyjelks: > we should be able gracefully handle the absence of any field on the client side

Agreed!

Since Myk is ooo the rest of the day and I'd like to try to move this work forward, does anyone object to removing the boolean field in a follow-up PR? We think this is safe to merge? cc: @mzikherman @anandaroop

ashleyjelks:
> Works for me as-is, with the understanding that we will not use the new boolean introduced here, as well as remove it from Kaws and MP in a follow-up

Sounds good, I'll make a ticket to track the follow-up work!

PR #188, opened 2020-01-14
Removes hasMarkdownDescription field from schema

ashleyjelks:
The hasMarkdownDescription boolean is not needed to support converting collections descriptions from HTML to markdown.

The PR removes this field made in #186.

#trivial


src/Entities/Collection.ts
```diff
@@ -44,9 +44,6 @@ export class Collection {
   @Field(type => Boolean, {
     description: "Collection has its description available as markdown",
   })
-  @Column({ default: false })
-  hasMarkdownDescription: boolean = false
-
```
anandaroop: Looks like we removed the db column (via the `@Column` decorator) but not the GraphQL field (via the `@Field` decorator)

src/Entities/Collection.ts
```diff
@@ -44,9 +44,6 @@ export class Collection {
   @Field(type => Boolean, {
```
mbilokonsky: everything from line 44 down is part of this, should be removed as well!

src/utils/__tests__/__snapshots__/createSchema.test.ts.snap
```diff
@@ -19,10 +19,7 @@ type Collection {
   descriptionMarkdown: String
 
   # Collection has its description available as markdown
-  hasMarkdownDescription: Boolean!
```
mbilokonsky: the commenting here is not quite right - we're deleting the correct comment and leaving the wrong one?

PR #190, opened 2020-01-14
Document Google credentials setup

anandaroop:
Hard-won lessons from https://artsyproduct.atlassian.net/browse/FX-1728

Wanted to document this further, while it's still fresh. 

(Also includes an unrelated schema bump from the markdown work happening currently (#186).

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -12,6 +12,9 @@ type Collection {
   # Description of the collection which can include links to other collections
   description: String
 
+  # Markdown alternate of description field contents.
+  descriptionMarkdown: String
+
   # Background image for the header of the collection page
   headerImage: String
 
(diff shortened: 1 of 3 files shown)


_schema.graphql
```diff
@@ -12,6 +12,9 @@ type Collection {
   # Description of the collection which can include links to other collections
   description: String
 
+  # Markdown alternate of description field contents.
```
mbilokonsky: do we need a rebase?

_schema.graphql
```diff
@@ -12,6 +12,9 @@ type Collection {
...
+  descriptionMarkdown: String
+
```
mbilokonsky: Is there a reason why this is in this PR? Shouldn't this already be in-place in kaws?

anandaroop: Nah, I think the `dump-schema` was just omitted from #186 (including it there would have been preferable, but not a dealbreaker since the dumped schema is mainly for MP's benefit, not Kaws')

I just snuck it in here since I'm in Kaws anyway

mbilokonsky: oooooh, I didn't check in that file - I assumed it was generated and got blown away. That makes sense, sorry about that!

PR #195, opened 2020-02-03
Enforce type-checking on PRs

pepopowitz:
Type-checking should be applied to all PRs. We currently run the `type-check` command as part of a git hook, so engineers can't push work that contains type errors, but dependabot can create PRs with broken types (because it doesn't use the git hook). 

We haven't had a case where dependabot submitted a PR to kaws that broke type-checking yet, but we did have one on positron ([related PR](https://github.com/artsy/positron/pull/2510)).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,9 @@
 version: 2.1
 
 orbs:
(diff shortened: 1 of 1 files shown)


sweir27:
Nice!! Thanks for adding this 🙏.

PR #196, opened 2020-02-06
[FX-1697] Limit number of related collections when querying MP for a collection

xtina-starr:
Addresses: [FX-1697](https://artsyproduct.atlassian.net/browse/FX-1697)

Currently we pull all the related collections (based on artist_ids or category). Since there are a large number of related collections we end up pulling more collections than necessary. This PR adds a `size` argument that allows us to specify the number of related collections we want to return.

related collections


src/Resolvers/Collections.ts
```diff
@@ -92,31 +92,42 @@ export class CollectionsResolver {
...
+    @Arg("size", () => Int, { nullable: true }) size: number,
     @Root() collection: Collection
   ): Promise<Collection[]> {
+    const query: { where: {}; take?: number } = { where: {} }
     const isArtistBased =
       collection.query.artist_ids && collection.query.artist_ids.length > 0
 
+    if (size) {
+      query.take = size + 1
+    }
```
xtina-starr: I use plus one here because a little further down we use `reject` to remove the initial collection queried from the related collection. This ensures that when we pass 6 as the `size` parameter that we get 6 related collections back.

src/Resolvers/Collections.ts
```diff
@@ -92,31 +92,42 @@ export class CollectionsResolver {
...
+        $in: collection.query.artist_ids,
+      }
+
+      const artistResults = await this.repository.find(query)
 
       const relatedArtistCollections = reject(artistResults, {
         id: collection.id,
       })
 
+      // We only want to return relatedArtistCollections when there are more
```
pepopowitz: This is a really helpful comment, thank you.

src/Resolvers/Collections.ts
xtina-starr: I initially thought to have a default value for the size property in the event a parameter wasn't passed but I thought this might be a less intuitive API because as a user I would expect if no `size` arg is explicitly passed to get back all the related collections that are available.

src/Resolvers/Collections.ts
```diff
@@ -92,31 +92,42 @@ export class CollectionsResolver {
...
   async relatedCollections(
+    @Arg("size", () => Int, { nullable: true }) size: number,
     @Root() collection: Collection
   ): Promise<Collection[]> {
+    const query: { where: {}; take?: number } = { where: {} }
     const isArtistBased =
       collection.query.artist_ids && collection.query.artist_ids.length > 0
 
+    if (size) {
+      query.take = size + 1
```
pepopowitz: I _think_ this is taking `size + 1` because we might end up filtering out the original collection from the results. Whether I'm right or wrong, I think we should add a comment here so readers aren't confused about why we're taking one extra.

src/Resolvers/Collections.ts
sweir27: Is there a use-case for needing to return _every_ related collection? That will suffer from scaling problems unless there are other restrictions on the number of possible related collections.

I think a sensible default would be best here (we use `10` in gravity)! If there are existing use-cases that need more collections, we can also update those and follow-up here by adding the default in a subsequent PR.

src/Resolvers/Collections.ts
```diff
@@ -92,31 +92,42 @@ export class CollectionsResolver {
...
-      },
-    })
+    query.where = {
+      category: { $in: [collection.category] },
+      show_on_editorial: true,
+    }
+
+    const relatedCategoryResults = await this.repository.find(query)
+
     const relatedCategoryCollections = reject(relatedCategoryResults, {
```
pepopowitz: I'm having a hard time describing this, so reach out via Slack if I'm not able to communicate well enough. 

I think we might need to add something like `.take(size)` to the end of this statement, in case the original collection was *not* included in the query results.

Example: I am working with collection 'photography'. I request 3 related collections, by sending `size: 3` as an argument. [We query for the first 4](https://github.com/artsy/kaws/pull/196/files#diff-500113cce4e1c0ce1050b25c61eb35a4R103) elements. We then [`reject` the 'photography' collection](https://github.com/artsy/kaws/pull/196/files#diff-500113cce4e1c0ce1050b25c61eb35a4R113) so that it's not included in its own "related" collections. But it turns out that none of the first 4 related collections were 'photography', so I still return 4 results, despite the client asking for only 3.

src/Resolvers/__tests__/collection_resolvers.test.ts
```diff
@@ -376,4 +446,36 @@ describe("Collection", () => {
...
+      }
+    `
+
+    return runQuery(query, {}, createMockSchema).then(data => {
+      expect(find).toBeCalledWith({
+        where: {
+          category: { $in: ["Pop Art"] },
+          show_on_editorial: true,
+        },
+        take: 3,
```
pepopowitz: Similar to an above comment, I think it would be useful to explain *why* we're expecting this call to request 1 more than we passed into the `size` argument.

src/Resolvers/Collections.ts
sweir27: This does feel a bit roundabout! Maybe instead we can limit to the size we want but explicitly exclude the current collection's ID from the results (via our query)?

src/Resolvers/Collections.ts
sweir27: ➕

src/Resolvers/Collections.ts
```diff
@@ -92,31 +92,42 @@ export class CollectionsResolver {
...
     }
 
-    const relatedCategoryResults = await this.repository.find({
-      where: {
-        category: { $in: [collection.category] },
-        show_on_editorial: true,
-      },
-    })
+    query.where = {
+      category: { $in: [collection.category] },
```
anandaroop: Seems like `category` is not array-valued, so I think this could simply be 

```suggestion
      category: collection.category,
```

src/Resolvers/Collections.ts
anandaroop: Oh tricky! 

I think this is another point in favor of Sarah's suggestion to simply use the Mongo query to filter out the originating collection, something like:

```js
query.where = {
  id: { $ne: collection.id},
  category: collection.category,
  show_on_editorial: true,
 }
```

(Please check my Mongo syntax 😅)

That avoids the +1 problem and sidesteps any need for post-filtering with `reject`.

src/Resolvers/Collections.ts
```diff
@@ -92,46 +92,38 @@ export class CollectionsResolver {
...
-        $in: collection.query.artist_ids,
-      }
-
-      const artistResults = await this.repository.find(query)
-
-      const relatedArtistCollections = reject(artistResults, {
-        id: collection.id,
+      const artistResults = await this.repository.find({
+        where: {
+          slug: { $ne: collection.slug },
```
xtina-starr: Went for the  `$ne` query method here as @anandaroop suggested. This lets us avoiding having to use `reject`.

src/Resolvers/Collections.ts
```diff
@@ -92,35 +92,38 @@ export class CollectionsResolver {
...
+      if (artistResults.length > 4) {
+        return artistResults
       }
     }
 
     const relatedCategoryResults = await this.repository.find({
       where: {
-        category: { $in: [collection.category] },
+        slug: { $ne: collection.slug },
+        category: { $in: collection.category },
```
anandaroop: Looking good 🎉  

But I'm wondering about this line — the `$in` should only be there if the test is against an array, which is not the case here. I'd expect the line above to cause an error, and instead to see something simpler like

```js
category: collection.category,
```

Then again, tests are green. So either I'm wrong or the tests are 😬

anandaroop: Looking closer at the tests, it looks like we [mock out the connection to the database](https://github.com/artsy/kaws/blob/83f412d2c40005691ab7985b5f9c5ccbfb4323b5/src/Resolvers/__tests__/collection_resolvers.test.ts#L8), so the tests wouldn't necessarily catch a malformed Mongo query.

If you'd prefer I'm happy to pair on this with you and confirm my suspicions 😄

xtina-starr: I went ahead and made the change but I'm down to pair in order to confirm.

src/Resolvers/Collections.ts
```diff
@@ -92,35 +92,38 @@ export class CollectionsResolver {
 
   @FieldResolver(type => [Collection])
   async relatedCollections(
+    @Arg("size", () => Int, { defaultValue: 10 }) size: number,
```
sweir27: 🙏 🙌

PR #198, opened 2020-02-13
FX-1801: Expose isDepartment in the API

sweir27:
Paired with @xtina-starr !

image

### Problem
Currently, client apps (currently just reaction, but soon Emission) have to do a little bit of business logic to determine whether or not a collection is a "department" or not. It's not _horribly_ messy, but given we're starting to treat departments as more first-class (i.e. by labeling them differently in search, potentially having a different URL structure), it feels like a good time to push more of this logic to the current source of truth (KAWS).

### Solution
We debated whether to have this be a field in the database or simply resolved at the API level. Given there's no explicit need just yet to have this be in the database (and that would necessitate additional changes, like changing our import script or the spreadsheet), it felt that we could achieve our current desired outcome by adding this field to the API. It should be easy enough to persist this value down the line if we need to.

### One Interesting Thing
Although the [logic in reaction](https://github.com/artsy/reaction/blob/1fd632de985d9a4cb58940091601c896da08031f/src/Apps/Collect2/Routes/Collection/index.tsx#L76) to determine whether a collection is a department or not only checks the length of the `linkedCollections` array, we chose to additionally verify that there is at least one `linkedCollection` _with `members`_.

It sort of suggests that down-the-line we'll want a more first-class understanding (vs. just deriving based on fields) of what a "department" is, but felt like an edge case to avoid for now. Let me know if you'd prefer we use the simpler logic for now!

### DIFF
--- a/src/Resolvers/__tests__/fixtures/data.ts
+++ b/src/Resolvers/__tests__/fixtures/data.ts
@@ -20,6 +20,7 @@ export const mockCollectionRepository = plainToClass(Collection, [
     show_on_editorial: false,
     price_guidance: null,
     is_featured_artist_content: true,
+    linkedCollections: [{ id: "abc", members: [{ name: "Linked 1" }] }],
   },
   {
     id: 2,
@@ -55,6 +56,7 @@ export const mockCollectionRepository = plainToClass(Collection, [
     show_on_editorial: false,
     price_guidance: 1000,
     is_featured_artist_content: true,
+    linkedCollections: [],
   },
   {
     id: 4,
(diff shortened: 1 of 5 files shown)


src/Resolvers/__tests__/fixtures/data.ts
```diff
@@ -20,6 +20,7 @@ export const mockCollectionRepository = plainToClass(Collection, [
     show_on_editorial: false,
     price_guidance: null,
     is_featured_artist_content: true,
+    linkedCollections: [{ id: "abc", members: [{ id: 3 }] }],
```
anandaroop: #minor Not a big deal, but here and elsewhere I was a little confused by the `id` attribute, since that's not part of the schema or current data (although `name` is). Maybe worth a cleanup to save us future confusion?

sweir27: Updated!

PR #199, opened 2020-02-14
Add package script to pull db from staging or production

anandaroop:
## Problem

Nowadays it has become expensive to run `yarn update-database` in order to bootstrap a local copy of realistic data (thanks to the price calculations that hit MP for most of the 2000+ collections)

## Solution

This PR adds a `yarn pull-database` script that will simply replace your local kaws database with the contents of the staging database. (You can also `yarn pull-database production` in order to pull down production data.)

pulldb

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
   ],
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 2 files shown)


sweir27:
Gonna merge this in so it's not sitting around! 🙏

PR #225, opened 2020-04-05
Allows for filtering collections by a list of slugs

sweir27:
As part of the "Art Keeps Going" campaign, we need to be able to pull a list of collections filtered by an arbitrary list of slugs (curated by an admin).

This seems like helpful functionality in any case!

Example of just querying by slugs:
image

Example of querying by slugs and another param (in this case `category`):
image

### DIFF
--- a/_schema.graphql
+++ b/_schema.graphql
@@ -141,6 +141,7 @@ enum GroupTypes {
 
 type Query {
   collections(
+    slugs: [String!]
     category: String
     randomizationSeed: String
     size: Int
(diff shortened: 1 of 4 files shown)


sweir27:
Going to merge this in so I can test dependent changes more easily, but happy to update tomorrow pending comments!

damassi:
Haha @sweir27 - this is just what I need to test my SSR collections rail thoughts I brought up on Thursday. Perfect timing!

PR #240, opened 2020-06-12
@mzikherman => [PriceGuidance] Correctly throw and exit from price guidance script

mzikherman:
Realized this script never actually ends if it ends without an error! :() h/t @dzucconi


src/utils/getPriceGuidance.ts
```diff
@@ -65,5 +65,6 @@ export const getPriceGuidance = async (slug: string) => {
     return hasNoBasePrice ? avgPrice : Math.ceil(avgPrice / 100) * 100
   } catch (error) {
     console.log(error)
+    throw error
```
mzikherman, line 68: Without throwing this, clients of this helper don't have the opportunity to do anything with the error. All uses of this method are _already_ inside of try/catch.

Overall, the number of try/catch (which in many cases logs the error but doesn't propagate) is pretty high in these methods.

PR #250, opened 2020-08-07
Add ingress resources (for migration from ELB) #migration

joeyAghion:
Set up ingress resources in both staging and production, as per doc:

https://www.notion.so/artsy/Nginx-Ingress-Controllers-4b5d937f0acd480f938c34ef1509d3c4

I already enabled Cloudflare proxying on both staging and production, since that's a worthwhile improvement anyway and we can combine those settings with this change.

Traffic won't actually use the new ingress service until DNS is updated to point to the shared "network" load balancer. A follow-up PR and migration will remove the unnecessary ELB service.

### Migration

Once deployed to either environment, update `kaws[-staging]`'s DNS to `<the network load-balancer specified in above doc>`.

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -144,6 +144,46 @@ spec:
     component: web
   sessionAffinity: None
   type: LoadBalancer
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: kaws
+    component: web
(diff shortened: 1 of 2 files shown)


joeyAghion:
Gonna go ahead and merge but feedback still welcome.

PR #251, opened 2020-08-11
Fix production ingress definition

joeyAghion:
Follow-up to https://github.com/artsy/kaws/pull/250.

This caused an issue in production, as you'd expect.


anandaroop:
Thanks! 😄

PR #264, opened 2020-09-24
feat: add a CLI for collection deletion

anandaroop:
Addresses [FX-2290]

### Problem

Kaws doesn't support deletion of collections, thus requiring developers to monkey around in a Mongo console. Even then, deleted collections remain indexed for search.

### Solution

Add an interface for developers to conveniently and correctly remove collections from Kaws.

My first thought was to add e.g.  a `deleteCollection` mutation to Kaws' GraphQL API. But then I realized we don't enforce any auth on the GraphQL endpoint, so that rules out doing destructive operations there.

Second thought was a CLI — so now this can be accomplished with simply a yarn script and some collection slugs:

```sh
$ yarn delete-collections alex-katz-portraits alex-katz-landscapes
```

Here's what a sample run of this CLI looks like:

Screen Shot 2020-09-24 at 4 45 02 PM

[FX-2290]: https://artsyproduct.atlassian.net/browse/FX-2290

### DIFF
--- a/scripts/deleteCollections.ts
+++ b/scripts/deleteCollections.ts
@@ -1,0 +1,14 @@
+import { CollectionRemover } from "../src/lib/utils/collectionRemover"
+
+const slugs = process.argv.slice(2)
+console.log(
+  `Attempting to remove ${slugs.length} collection(s) from Elasticsearch and MongoDB`
+)
+
+const collectionRemover = new CollectionRemover(slugs)
+
+collectionRemover.setup().then(() => {
+  collectionRemover.perform().finally(() => {
+    collectionRemover.teardown()
+  })
+})
--- a/src/lib/utils/collectionRemover.ts
+++ b/src/lib/utils/collectionRemover.ts
@@ -1,0 +1,101 @@
+import {
+  Connection,
+  createConnection,
+  getMongoRepository,
+  MongoRepository,
+} from "typeorm"
+import { Client as ElasticsearchClient } from "elasticsearch"
+import { red, yellow, green } from "bash-color"
+
+import { databaseConfig } from "../../config/database"
+import { Collection } from "../../Entities/Collection"
+
+export class CollectionRemover {
+  private slugs: string[]
+  private mongoConnection: Connection
+  private mongoRepository: MongoRepository<Collection>
+  private elasticsearchClient: ElasticsearchClient
+  private elasticsearchIndexName: string =
+    "marketing_collections_" + process.env.NODE_ENV
+
+  constructor(collectionSlugs: string[]) {
+    this.slugs = collectionSlugs
+    this.checkEnvironment()
+  }
+
+  checkEnvironment = () => {
+    const { MONGOHQ_URL, ELASTICSEARCH_URL } = process.env
+    if (!MONGOHQ_URL) throw new Error("MONGOHQ_URL must be defined")
+    if (!ELASTICSEARCH_URL) throw new Error("ELASTICSEARCH_URL must be defined")
+  }
+
+  setup = async () => {
+    // setup Mongo
+    this.mongoConnection = await createConnection(databaseConfig())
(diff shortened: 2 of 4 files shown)


mzikherman:
> My first thought was to add e.g. a deleteCollection mutation to Kaws' GraphQL API. But then I realized we don't enforce any auth on the GraphQL endpoint, so that rules out doing destructive operations there.

I'm not sure if a mutation is the best choice, just because there would still be the issue of the ElasticSearch delete. But, if that ES delete could be wrapped up in the mutation that _might_ be preferable than a custom script? In terms of, possibly being more easily scriptable from outside KAWS (not requiring a console).

Re: authentication, I think that's intentional. KAWS is an internal app and is on the VPC, so can only be accessed over the VPN (and thus _can_ be open).

anandaroop:
Yeah my thought was the mutation would take care of any side effects, including ES deletion. 

But the services (and the gql playground) _are_ presently accessible without connecting to the VPN, no?

mzikherman:
> But the services (and the gql playground) are presently accessible without connecting to the VPN, no?

Oh yea, of course :) I don't know why I thought that!

anandaroop:
I mean, that could be worth changing too 🤷

scripts/deleteCollections.ts
```diff
@@ -0,0 +1,14 @@
...
+  `Attempting to remove ${slugs.length} collection(s) from Elasticsearch and MongoDB`
+)
+
+const collectionRemover = new CollectionRemover(slugs)
+
+collectionRemover.setup().then(() => {
+  collectionRemover.perform().finally(() => {
+    collectionRemover.teardown()
+  })
+})
```
anandaroop, line 14: Until we get top-level async/await, a good old fashioned pyramid of doom 😄

src/lib/utils/collectionRemover.ts
```diff
@@ -0,0 +1,101 @@
...
+import { databaseConfig } from "../../config/database"
+import { Collection } from "../../Entities/Collection"
+
+export class CollectionRemover {
+  private slugs: string[]
+  private mongoConnection: Connection
+  private mongoRepository: MongoRepository<Collection>
+  private elasticsearchClient: ElasticsearchClient
+  private elasticsearchIndexName: string =
+    "marketing_collections_" + process.env.NODE_ENV
```
anandaroop, line 19: Because of our env-specific index naming

src/lib/utils/collectionRemover.ts
```diff
@@ -0,0 +1,101 @@
...
+  perform = async () => {
+    await Promise.all(this.slugs.map((slug) => this.removeCollection(slug)))
+  }
+
+  removeCollection = async (slug: string) => {
+    try {
+      const collection = await this.mongoRepository.findOne({ slug })
+      if (collection) {
+        await this.removeFromSearchIndex(collection!)
+        await this.removeFromDatabase(collection!)
```
anandaroop, line 56: This is what it's all about.

src/lib/utils/collectionRemover.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  removeFromSearchIndex = async (collection: Collection) => {
+    try {
+      await this.elasticsearchClient.delete({
+        index: this.elasticsearchIndexName,
+        type: "marketing_collection",
+        id: collection.id.toString(),
+      })
+    } catch (e) {
+      if (e.message === "Not Found") {
```
sweir27, line 75: Oooh nice-- so if we end up with a situation where (for some reason) we're able to delete from the search index but _not_ from the database and we re-run... we'll end up with this helpful warning instead of a failure! I like it.

sweir27 (approved):
Nice!!!

anandaroop:
Description updated with instructions for running on staging/prod.

PR #272, opened 2020-11-17
Prepare for internal routing #migration

joeyAghion:
For [PLATFORM-1312](https://artsyproduct.atlassian.net/browse/PLATFORM-1312), we want to stop exposing Kaws's API publicly. It's used exclusively by other systems within the cluster, and occasionally by developers during development, all of which can be accomplished without a public API (with the potential vulnerabilities and unqualified traffic that that entails).

Like in https://github.com/artsy/osmosis/pull/117 and https://github.com/artsy/delta/pull/136, clients within the cluster will be able to reach Kaws at `kaws-web-internal:8080` and developers on the VPN can use `kaws.[stg|prd].artsy.systems`. Unlike those PRs, we'll migrate in 2 steps. This PR drops SSL-redirecting functionality and exposes the new URLs. After clients are updated, a follow-up PR will drop the public ingress definitions.

### Pre-deploy:
* [x] Create new `kaws.[stg|prd].artsy.systems` DNS records in Route53 (pointing at `nginx.[stg|prd].artsy.systems`)

### Post-deploy:
* [ ] Test direct access from any `hokusai staging run --tty bash`-type console: `curl -vvv -H 'Content-Type: application/json' http://kaws-web-internal:8080/graphql -d '{"query":"{collections{slug}}"}'`
* [ ] update Metaphysics' config: `hokusai [staging|production] env set KAWS_API_BASE=http://kaws-web-internal:8080; hokusai [staging|production] refresh`
* [ ] update Gravity's config `hokusai [staging|production] env set KAWS_API_URL=http://kaws-web-internal:8080/graphql; hokusai [staging|production] refresh`
* [ ] update Metaphysics' `.env.shared` file with the new `kaws.stg.artsy.systems` URL and re-upload: `aws s3 cp .env.shared s3://artsy-citadel/dev/.env.metaphysics`
* [ ] delete `kaws[-staging].artsy.net` DNS in Cloudflare
* [ ] update [Notion] project list with new URLs
* [ ] follow-up PR: drop unused public ingress definitions from `.yml` files (since this won't delete them from the clusters, that will have to be done manually)

*Please preserve commits. They're distinct changes but should be released together.


.env.example
```diff
@@ -1,6 +1,5 @@
 NODE_ENV=development
 PORT=4000
-APP_NAME=kaws-staging
```
joeyAghion, line 3: As far as I can tell, this was only used by the `ensureSSL` middleware, as far as I can tell.

README.md
```diff
@@ -8,17 +8,17 @@ collections you ask? A Collection is a prefiltered version of Artsy's
...
-  [k8s](https://kubernetes-staging.artsy.net/#!/search?q=kaws&namespace=default)
+  [http://kaws.stg.artsy.systems](http://kaws.stg.artsy.systems/playground) |
+  [k8s](https://kubernetes-staging.artsy.net/#!/search?q=kaws&namespace=default) or `kaws-web-internal:8080` within the cluster
 - **GitHub:** [https://github.com/artsy/kaws/](https://github.com/artsy/kaws/)
 - **CI:** [CircleCI](https://circleci.com/gh/artsy/kaws); merged PRs to
   artsy/kaws#master are automatically deployed to staging. PRs from `staging` to
   `release` are automatically deployed to production.
   [Start a deploy...](https://github.com/artsy/kaws/compare/release...staging?expand=1)
-- **Point People:** [@mbilokonsky](https://github.com/mbilokonsky),
+- **Point People:** [@anandaroop](https://github.com/anandaroop),
```
joeyAghion, line 21: Objections? 😬

hokusai/production.yml
```diff
@@ -121,6 +121,23 @@ spec:
               serviceName: kaws-web-internal
               servicePort: http
 
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: kaws-internal
```
joeyAghion, line 128: Conventionally we'd just call this `kaws`, but that's occupied by the public ingress defined above. In the follow-up PR dropping that, I'll probably update this name.

hokusai/production.yml
```diff
@@ -199,7 +216,7 @@ kind: CronJob
 metadata:
   name: kaws-update-site-map
 spec:
-  schedule: 7 15 * * *
+  schedule: 7 16 * * *
```
joeyAghion, line 219: Unrelated change (sorry): I noticed that Kaws schedules 2 different crons for the same moment.

README.md
anandaroop: Heh, sure let's go with that 😃

hokusai/production.yml
```diff
@@ -121,6 +121,23 @@ spec:
...
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: kaws-internal
+  annotations:
+    kubernetes.io/ingress.class: nginx-internal
+spec:
+  rules:
+    - host: kaws.prd.artsy.systems
```
anandaroop, line 133: So this (and `stg`) is our new url for dev faffing 👍

src/index.ts
```diff
@@ -1,6 +1,5 @@
 import "dotenv/config"
 import { setupDataDog } from "./config/datadog"
-import { ensureSSL } from "./lib/ensureSSL"
```
anandaroop, line 3: Is it simply unnecessary, or actually an anti-pattern to use SSL for these within-cluster connections?

anandaroop (approved):
Thanks!

src/index.ts
joeyAghion: It's hopefully unnecessary, given how inter-host traffic within the Virtual Private Cloud is already encrypted. But it's not an anti-pattern: the old worries about encryption/decryption performance are no longer a significant concern, according to my understanding. Most importantly: we don't currently _support_ SSL traffic to the internal nginx deployment. That will be addressed by [PLATFORM-2981](https://artsyproduct.atlassian.net/browse/PLATFORM-2981), after which SSL should work internally as well.

PR #274, opened 2020-11-18
Remove public ingress declaration

joeyAghion:
This is the follow-up to https://github.com/artsy/kaws/pull/272. DNS for the public URLs has already been deleted, and logs confirm no traffic through the external ingress.

Question: Just to stick with our naming convention, I've renamed the internal ingress to `kaws` and deleted the external one with the same name. When these changes are applied, I assume the former `kaws-internal` definition will remain and require manual deletion. Is it a problem that 2 ingress definitions will temporarily list the same host? Any reason to worry that this update won't be graceful?

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -110,23 +110,6 @@ kind: Ingress
 metadata:
   name: kaws
   annotations:
-    nginx.ingress.kubernetes.io/whitelist-source-range: "{{ cloudflareIpSourceRanges|join(',') }}"
-spec:
-  rules:
-    - host: kaws.artsy.net
-      http:
-        paths:
-          - path: /
-            backend:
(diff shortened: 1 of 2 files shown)


anandaroop (approved):

artsyjian:
Agreed that we should stick with `kaws` for ingress name. I sort of remember testing this (two ingresses with the same `host` field). I think ingress controller won't complain, both will exist. Both will go into Nginx config, and it will just work. `kaws-internal` can be deleted subsequently w/o disruption.

artsyjian:
Ah. But in this case, the old ingress has an ip allowed-list. Curious to see what the behavior will be.

joeyAghion:
Deleted the orphaned `kaws-internal` ingress from staging.

PR #277, opened 2020-11-18
Replace /upload endpoint with import-collections command

joeyAghion:
Upon removing public DNS for Kaws as part of https://github.com/artsy/kaws/pull/272, it was reported that syncing from Google spreadsheets raised errors. This makes sense (and was an oversight), since the spreadsheet integration can no longer reach Kaws's public `/upload` endpoint.

This PR adds a `yarn import-collections` command and drops the `/upload` endpoint. Engineers can invoke it via `hokusai production run ...`, but our plan is to set up an [on-demand Jenkins job](https://joe.artsy.net/job/kaws-import-collections) where admins can press a "build now" button to accomplish this directly.

Co-authored by: @anandaroop


anandaroop (approved):
👍

PR #306, opened 2021-04-16
feat: protect cronjob from eviction by cluster-autoscaler, and true up with hokusai template.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3202

- Set [safe-to-evict](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) to false for crons so they cannot be evicted by cluster-autoscaler.
- Generally, true up with hokusai [template](https://github.com/artsy/artsy-hokusai-templates/tree/master/rails-puma/hokusai).

for peril:
#trivial

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -31,6 +31,8 @@ spec:
           env:
             - name: PORT
               value: "8080"
+            - name: NODE_ENV
+              value: production
             - name: DD_TRACE_AGENT_HOSTNAME
               valueFrom:
                 fieldRef:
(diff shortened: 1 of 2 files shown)


hokusai/production.yml
```diff
@@ -31,6 +31,8 @@ spec:
           env:
             - name: PORT
               value: "8080"
+            - name: NODE_ENV
+              value: production
```
artsyjian, line 35: should be safe?

hokusai/production.yml
```diff
@@ -284,7 +298,11 @@ spec:
   concurrencyPolicy: Forbid
   jobTemplate:
     spec:
+      backoffLimit: 2
```
artsyjian, line 301: not sure why only this cronjob doesn't have backoffLimit set.

anandaroop (approved):

PR #418, opened 2022-04-21
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)

artsyjian:
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)
#trivial


ovasdi:
#trivial

ovasdi:
@artsyjian type-check failure due to node version incompatibility with yarn orb which now depends on [node 14](https://github.com/artsy/orbs/commit/5cdcdd95fb7b927127f5f8b9a4efec4216fd8a27)

